Issue #29, opened 2025-12-22
Python Module virtualenv not being created and/or not picking up the `requirements.txt` file


caenrigen:
@ximion following up from our discussion in #25

Info:
Syntalos installed via Flatpak, updated to the latest version after you shipped it with `pyserial` built in.

```bash
$ uname -a
Linux debianvm 6.17.12+deb14-arm64 #1 SMP PREEMPT Debian 6.17.12-1 (2025-12-14) aarch64 GNU/Linux
```

`Diagnostics -> System`:
```
Syntalos 2.1.1 | Debian GNU/Linux forky/sid (Kernel: linux 6.17.12+deb14-arm64)
Software: Qt 6.10.1 | OpenCV 4.12.0 | FFMpeg 7.1.3 | Python 3.13
Sandbox: Flatpak; Runtime: KDE Flatpak runtime 6.10
Architecture: arm64 | CPU:  | OpenGL: 4.5 (Compatibility Profile) Mesa 25.2.6 (git-3f12436d80)
Misc: Missing hardware support package.
```

---

After I upgraded Syntalos, the first time I tried to use my custom Python Module (via symlink, though probably this is not relevant) a terminal window opened and stayed opened with the following errors:

```
- virtualenv .
created virtual environment CPython3.13.9.final.0-64 in 186ms
  creator CPython3Posix(dest=/home/victor/.var/app/org.syntalos.syntalos/data/venv/shimmer, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, via=copy, app_data_dir=/home/victor/.local/share/virtualenv)
    added seed packages: pip==25.3
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
- source /home/victor/.var/app/org.syntalos.syntalos/data/venv/shimmer/bin/activate
- pip install -r /run/user/1000/shimmer-requirements_tlEy.txt
ERROR: Could not open requirements file: [Errno 2] No such file or directory: '/run/user/1000/shimmer-requirements_tlEy.txt'

Command failed to run. Press enter to exit.
```

When this window opens and stay opens the virtual env dir `/home/victor/.var/app/org.syntalos.syntalos/data/venv/shimmer` actually gets some files created inside it.

[...]

PR #32, opened 2025-12-23
Wait for external terminal to finish executing the virtualenv creation

caenrigen:
This prevents the temporary requirements.txt file being deleted before `pip install -r ...` has time to use it

Fixes (partially) #29

### DIFF
--- a/src/fabric/executils.cpp
+++ b/src/fabric/executils.cpp
@@ -25,6 +25,7 @@
 #include <QFileInfo>
 #include <QProcess>
 #include <QStandardPaths>
+#include <QThread>
 #include <glib.h>
 
 #include "sysinfo.h"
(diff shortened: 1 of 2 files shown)


ximion:
This looks weird - the terminal shouldn't detach... (and I can't reproduce this here)
I wonder if this happens if the terminal is already open? (in which case the command would complete immediately)

Now that we depend on QTermWidget anyway, I am quite tempted to drop all that brittle logic entirely, and just use an internal terminal emulator instead of any external one.

ximion:
> That sounds indeed like a more robust solution. When I checked the code I was wondering why did you decide to invoke external application, doesn't C++ (or Qt) allow to execute system calls directly? For UX a simple text window showing the output should be enough for debugging purposes

You need a proper terminal emulator, because some of these scripts do nasty interactive stuff, or render animations and color. And those are not trivial to implement at all, plus, opening the familiar terminal seemed like a good idea at the time.

Back then, no QTermWidget dependency existed yet and was not desirable - but now that we depend on it, it would be a lot more stable to ditch external terminals in favour of using an embedded one. I just need to find some time to write that code :-)

caenrigen:
I think I will just keep `konsole` installed for now to avoid having to deal with `gnome-terminal`, like that I will not need this PR to be merged anytime soon

Issue #41, opened 2026-02-10
Problems with Basler USB3 camera via Aravis module


caenrigen:
Hi @ximion 

I am having some pretty unstable experience with Basler USB3 camera via the Aravis module.
It seems to be mainly due to the saving/loading of settings, but it is a big issue because it kills reproducibility and the stability of the setup, and occasionally it crashes Syntalos.

Full disclosure: I don't have yet much experience using Basler cameras or any similar ones, nor the associated libraries for configuring and acquiring frames. I hope your experience can shed some light on how to achieve a reliable setup.

I am changing just the FPS (=60), Gain and Exposure. I am not touching any advanced settings (should I?).

I am seeing similar issues with both a Basler a2A1920-160uc and a Basler acA640-750um. The two cameras share the same USB HostController, the only one on this machine that is a high speed, in principle this I think is not a problem once the cameras are streaming.

---

### The problem(s)

The first time I create a project and add a camera things work well. But when I save and load the `.syct` project again (without rebooting the machine nor removing any of the hardware), then I often get errors like these (example for one of the two Basler cameras):

```
camera-arv-1: [10:44:26] Found 3 cameras.
camera-arv-1: [10:44:26] Pixel formats: QList("Mono8", "Mono10", "Mono10p", "Mono12", "Mono12p", "RGB8", "BGR8", "YCbCr422_8", "BayerRG8", "BayerRG10", "BayerRG10p", "BayerRG12", "BayerRG12p") 
camera-arv-1: [10:44:26] Network address not found, trying best-effort MTU 1500.
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
camera-arv-1: [10:44:26] Incompatible camera settings
[...]

PR #47, opened 2026-02-13
Fix uninitialized m_realFps

caenrigen:
This is a follow up from #41 

The reason my syntalos gets "sporadically" `Killed` is due to a buggy control flow and an uninitialised `m_realFps` in some cases. I have video that show the problem in more detail if needed, but the gist is that `m_realFps` has some random massive value from garbage memory and when Syntalos uses it it gets clapped to the max allowed by the camera (the `1000000` value) which eventually spikes memory allocation attempt and leads to the `Killed`.

NB AI-generated solution. It seemed to fix the bug. Tweak as needed

---

### How I reproduce the issue (sometimes)

I save a board with e.g. 60 FPS, close Syntalos and basler_aca640_gain_probe.c (Codex-written tool that allows to read/write Aravis settings directly outside Syntalos) reports `AcquisitionFrameRate: 60.0024000960 Hz`.

Then I open syntalos again change FPS to 61 FPS, do NOT save the board, close syntalos. probe reports `AcquisitionFrameRate: 61.0016470445 Hz`.

Open syntalos and I get the error pop up:
```
camera-arv-1: [09:39:51] Setting failure, wanted: "\tAcquisitionFrameRate\tFloat\t60.002400096003839\tHz" 
camera-arv-1: [09:39:51] actual: "\tAcquisitionFrameRate\tFloat\t61.001647044470204\tHz" 
```

I dismiss it and start an ephemeral run, all good camera working, I stop the run, all good so far. I try running a second ephemeral run and syntalos gets `Killed`. Probe reports `AcquisitionFrameRate: 1000000.00 Hz`

## What Codex identified:

- m_realFps is used uninitialized in the external run stop path, which can write garbage FPS (often clamped to max).
- Declared but not initialized: configwindow.h (line 131)
- Constructor does not initialize it: configwindow.cpp (line 45)
- External stop path calls preview-stop logic and restores camera->setFPS(m_realFps): configwindow.cpp (line 511) and configwindow.cpp (line 514)
- Immediately after, setCameraInUse(false) re-applies fpsSpinbox->value(), which can already have been updated to the bad value via dataChanged: configwindow.cpp (line 460)

## What else Codex seems to have spotted:

Advanced feature restore parser is structurally incompatible with serialized format.
- Serializer writes category lines as Category: ...: qarvfeaturetree.cpp (line 110)
- Parser only skips Category (no colon): qarvcamera.cpp (line 576)
- Serializer writes float units (Hz, us, dB): qarvfeaturetree.cpp (line 146)
- Parser reads only 3 whitespace tokens (name type value), so unit tokens desynchronize parsing: qarvcamera.cpp (line 581)

I have no idea if there is a real issue regarding this or not.

---

I am still getting the FPS setting warning, which I would expect to not happen, even if camera settings have been changed externally or during syntalos previous run (without saving the board) it should just override the setting and not complain. The setting in the board should be the source of truth. Perhaps this is the quirk of GenICams you have been mentioning, or perhaps related to all of the above.

### DIFF
--- a/modules/camera-arv/configwindow.cpp
+++ b/modules/camera-arv/configwindow.cpp
@@ -45,6 +45,7 @@ using namespace QArv;
 ArvConfigWindow::ArvConfigWindow(const QString &modId, QWidget *parent)
     : QMainWindow(parent),
       m_modId(modId),
+      m_previewFpsCaptured(false),
       camera(nullptr),
       decoder(nullptr),
       playing(false),
@@ -504,6 +505,7 @@ void ArvConfigWindow::toggleVideoPreview(bool start)
 
             // we only record with a low framerate for the preview
             m_realFps = camera->getFPS();
+            m_previewFpsCaptured = true;
             camera->setFPS(10);
             pixelFormatSelector->setEnabled(false);
             camera->startAcquisition();
@@ -511,7 +513,11 @@ void ArvConfigWindow::toggleVideoPreview(bool start)
     } else if (!start && started) {
         started = false;
         camera->stopAcquisition();
-        camera->setFPS(m_realFps);
+
+        // External run teardown also lands here; only restore preview baseline if it was captured.
+        if (m_previewFpsCaptured)
+            camera->setFPS(m_realFps);
+        m_previewFpsCaptured = false;
         decoder.reset();
 
         setCameraInUse(false);
--- a/modules/camera-arv/configwindow.h
+++ b/modules/camera-arv/configwindow.h
@@ -129,6 +129,7 @@ private:
     QString m_modId;
     QMetaObject::Connection m_debugConnection;
     int m_realFps;
+    bool m_previewFpsCaptured;
     std::shared_ptr<QArvCamera> camera;
     std::shared_ptr<QArvDecoder> decoder;
 


caenrigen:
Uninitialised variables are really nasty, in this case if the trash memory does not contain some big number, the bug does not surface (hence my trouble with hunting it down and sometimes it randomly seemed to work and not get `Killed`), unless you go into the Aravis settings and spot FPS changed to e.g. some random `20` (caught one of these on video today). Of course a total killer for reproducible experiments 🥲

ximion:
The AI explanation is weird in this case - it has a lot of good points, but the logic doesn't make sense...
The `m_realFps` value can never be non-initialized, because it only is read if `started` is also *true*. And that is only the case if `setCameraInUse(true)` was called. And that always happens in conjunction with setting `m_realFps`.

The only way this can be broken is if you run the board first, which triggers `setCameraInUseExternal`, and then while the camera is running open settings, then stop the run, and then close the settings. However, in that case, `toggleVideoPreview` isn't called since `playButton->isEnabled()` is false, so again the issue can't happen.

The only way this can possibly cause an issue would be if there was some weird race condition. All that said, I don't like that there is a possibility at all for this to happen by a spurious call the the GUI event loop processing, so, that has to be fixed.

I have not managed to reproduce the issue, but maybe the patch that closes this fixes the issue for you?
The fact that your way of reproducing this includes a save/load step makes me think that this problem is elsewhere...

caenrigen:
> I have not managed to reproduce the issue, but maybe the patch that closes this fixes the issue for you? The fact that your way of reproducing this includes a save/load step makes me think that this problem is elsewhere...

To be sure I am not messing up anything in my explanation of how the bug is reproduced I captured it on video:

https://youtu.be/y_jH4IBunTs

---

I tried reproducing it again a few times using your latest c2e135c76563c591f0b02b4172fd85226226db7e and this bug seems solved 🥳

PR #49, opened 2026-02-13
[Draft] camera-arv: Fix the parser incompatibility in restore logic.

caenrigen:
This seems to fix the issue I mentioned in https://github.com/syntalos/syntalos/pull/47#issue-3936420369 :

> I am still getting the FPS setting warning, which I would expect to not happen, even if camera settings have been changed externally or during syntalos previous run (without saving the board) it should just override the setting and not complain. The setting in the board should be the source of truth. Perhaps this is the quirk of GenICams you have been mentioning, or perhaps related to all of the above.

In fact if I change e.g. gain and exposure during one syntalos session without saving the board, and then open the board again I get the warning about all the modified settings

NB AI-suggested fix again, maybe helps you understand the #47 better too

---

## What changed (according to Codex):

- Reworked advanced-feature restore parser to read line-by-line and parse tab-delimited fields (matching serializer output) in qarvcamera.cpp (line 573)
- Explicitly skips serialized category lines (Category: ...) instead of token-based skipping.
- Handles feature rows as: `<name>\t<type>\t<value>[\t<unit>]`
- For Float, now ignores optional unit column and uses only the numeric value column.
- Added null/type guards before applying values to GenICam nodes.
Register restore now uses serialized hex payload (0x...) instead of the old broken token behavior.

### DIFF
--- a/modules/camera-arv/qarv/qarvcamera.cpp
+++ b/modules/camera-arv/qarv/qarvcamera.cpp
@@ -571,34 +571,57 @@ QTextStream& operator>>(QTextStream& in, QArvCamera* camera) {
     }
 
     while (!in.atEnd()) {
-        QString name, type, v;
-        in >> name;
-        if (name == "Category") continue;
-        ArvGcFeatureNode* node =
-            ARV_GC_FEATURE_NODE(arv_gc_get_node(camera->genicam,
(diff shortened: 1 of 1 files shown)


caenrigen:
Working on it :)

PR #56, opened 2026-03-02
engine: Unsubscribe dormant modules

### LINKED ISSUE #55: Memory explodes when streaming data from a Python Module into a disabled Module
@ximion run into a nasty one: memory keeps growing continuously when I disable a module that has incoming data from a Python Module. I might have experienced this problem before but there was too much going on to understand that this was the source.

Seems reproducible. The output module does not seem to matter. A fake Aravis input does not seem to cause anything like this.

Codex says this:

> The bug is in Syntalos' subscription lifecycle for disabled modules.
> 
> A disabled sink keeps its input subscription attached, but never consumes it.
> Then MLinkModule still forwards Python output because it sees “has subscribers”, and data accumulates in an effectively unbounded per-subscription queue.
> 
> Where this happens
> 
> Subscriptions are created regardless of module enabled/disabled state:
> 
> modulegraphform.cpp:387
> mainwindow.cpp:950
> 
> Disabled modules are put in DORMANT and not started (so they never drain inputs):
> 
> engine.cpp:1658
> engine.cpp:1784
> engine.cpp:1972
> 
> Python/MLink source still forwards when any subscriber exists:
> 
> mlinkmodule.cpp:833
> mlinkmodule.cpp:848
> mlinkmodule.cpp:814
> 
> Forwarded frames are pushed into subscription queues even if sink is dormant:
> 
> stream.h:537
> stream.h:391
> 
> Queue is not hard-capped; it allocates new blocks when full:
> 
> readerwriterqueue.h:602
> readerwriterqueue.h:631
> 
> So the concrete bug: disabled-module subscriptions are left active (not suspended/unsubscribed), and the queue implementation can grow, causing RAM blow-up when upstream keeps producing (Python modules make this very visible).

Testing a fix, draft PR in a sec

caenrigen:
Fixes #55 

AI-generated, change as needed

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -1786,6 +1786,26 @@ bool Engine::runInternal(const QString &exportDirPath)
         mod->setState(ModuleState::DORMANT);
     }
 
+    // Disabled or dormant modules do not actively process input, so their
+    // subscription queues must be suspended to avoid unbounded buffering.
+    std::vector<std::shared_ptr<VariantStreamSubscription>> suspendedInputSubscriptions;
+    auto suspendInputsForModule = [&suspendedInputSubscriptions](AbstractModule *mod) {
+        for (const auto &iport : mod->inPorts()) {
+            if (!iport->hasSubscription())
+                continue;
+            auto sub = iport->subscriptionVar();
+            sub->suspend();
+            suspendedInputSubscriptions.push_back(std::move(sub));
+        }
+    };
+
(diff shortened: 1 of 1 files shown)


ximion:
This was already pretty good, but I implemented a slightly different solution to ensure all modules are always in a known state (with all inputs enabled) at the start of an experiment, regardless of what happened before.

Issue #61, opened 2026-03-04
Memory grows continuously (allocator spiral of doom)


caenrigen:
Run a few Ephemeral runs with this board (without closing Syntalos ofc) and you will see a stead increase in RAM use for the `syntalos` process:

<img width="1273" height="800" alt="Image" src="https://github.com/user-attachments/assets/f7fc407d-35c3-45a1-8e75-4e57d4030970" />

I tried running it with `valgrind` but it crashes when stopping the run:

[valgrind.log](https://github.com/user-attachments/files/25740069/valgrind.log)
[syntalos_debug_stdout_video_recorder_av1.log](https://github.com/user-attachments/files/25740070/syntalos_debug_stdout_video_recorder_av1.log)

caenrigen:
NB this is a follow up from #60

caenrigen:
Giving up on debugging this further, I lack the necessary expertise, will wait for your judgement.

I will use VP9 for now

ximion:
Which version of `libsvtav1` are you using on your system?

Btw, obviously don't run with those environment variables - they may reduce memory usage, but at a fairly huge performance cost (by avoiding some memory pools).

ximion:
> but syntalos is installed via `ninja install`

Ooooooh, maybe that explains it! Can you run `objdump -x ./src/syntalos |grep 'R.*PATH'` on the Syntalos binary and tell me the output? (adjust the path to the right binary, of course).

PR #64, opened 2026-03-04
More attempts for fixing "memory leaks"

caenrigen:
One more attempt at fixing #61 

~This builds on top of #63 PR~

Codex-generated

NB did NOT seems to solve the issue.

---

Codex said regarding one of the commits:

> Replaced manual malloc-based frame buffer allocation in vw_alloc_frame(..., allocate=true) with av_frame_get_buffer(aframe, 32).
> Kept behavior for allocate=false unchanged.
> 
> Previously, the buffer allocated via malloc was not owned by FFmpeg refcounted buffers, so av_frame_free() did not reclaim it.
> av_frame_get_buffer() attaches owned buffers to AVFrame, so teardown via av_frame_free() releases memory correctly across runs.

### DIFF
--- a/modules/videorecorder/videowriter.cpp
+++ b/modules/videorecorder/videowriter.cpp
@@ -512,7 +512,6 @@ VideoWriter::~VideoWriter()
 static AVFrame *vw_alloc_frame(int pix_fmt, int width, int height, bool allocate)
 {
     AVFrame *aframe;
-    uint8_t *aframe_buf;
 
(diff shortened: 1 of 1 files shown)


ximion:
The AI definitely forgot to attach any memory at all to the frame - which is funny, because that of course fixes the leak, but also makes the software unusable.
I didn't like this entire function, so I implemented a solution that is more in line with how FFmpeg is used today, and not 6 years ago.

PR #65, opened 2026-03-05
Unref codecopts, cst, swsctx

caenrigen:
Moved these little changes into this PR to leave #63 with only the packet-related changes that I would be more skeptical about merging


ximion:
We should probably reject cropping to 1x1 pixel - most codecs can't work with that, and there is no circumstance where you would ever want that or record a single pixel (unless you have a microscope with a PMT, but then you wouldn't pass that data around as cv::Mat :-P).

Issue #43, opened 2026-02-11
Aravis Camera Module Region of Interest GUI broken and crashes Syntalos (segmentation fault)


caenrigen:
@ximion as I mentioned in #41 the ROI GUI has some bugs that make it unusable, in many cases it messes up the camera streaming and in such cases attempting to run Syntalos will lead to a segmentation fault.

[SyntalosCrashReport_261102_3.md](https://github.com/user-attachments/files/25233892/SyntalosCrashReport_261102_3.md) this one is the latest corresponding to the video

The following two I believe are also this same bug (but not 100% sure)
[SyntalosCrashReport_261102.md](https://github.com/user-attachments/files/25233893/SyntalosCrashReport_261102.md)
[SyntalosCrashReport_261102_2.md](https://github.com/user-attachments/files/25233894/SyntalosCrashReport_261102_2.md)

Video: https://youtu.be/KtKRdr5zsWY

As a general observation, the GUI becomes unresponsive often while attempting to change any of the ROI settings

Funny enough, in the video I captured the only time that I was able to actually get a ROI to be displayed correctly (the small region around the LED), most of the time I get the strange stripes/noise (+ Syntalos crash when attempting an acquisition)

PR #72, opened 2026-03-13
videotransform: Validate received frames

caenrigen:
Before and after this patch: https://youtu.be/oKkCU7fX9J0

@ximion I was debugging something else and then run into this crash.

Codex says the root cause is in the Aravis module itself, metadata goes out of sync with the decoder. Perhaps related to #43.

On top of that the Video Transformer is trusting the metadata and crashes Syntalos when the crop does not make sense anymore.

I would say that generally modules should not trust input frames "blindly" and validate against the "expected data format contract", this PR does this validation for the Video Transformer.

Maybe there are use cases/modules in which dynamic frame size is a desired feature, but at least for a fix-position-and-size cropping it makes sense to apply the validation.

---

I think a similar issue is affecting the Canvas when the Aravis changes, e.g., from colour to mono

### DIFF
--- a/modules/videotransform/videotransformmodule.cpp
+++ b/modules/videotransform/videotransformmodule.cpp
@@ -35,6 +35,7 @@ private:
 
     VTransformCtlDialog *m_settingsDlg;
     QList<std::shared_ptr<VideoTransform>> m_activeVTFList;
+    QSize m_expectedInputSize;
 
 public:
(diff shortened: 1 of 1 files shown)


ximion:
Hmm... Modules generally trust the data they receive, simply because validation can be very expensive and measurably slows down the system.

In your case, is the camera really changing the dimensions of the output frames during a run? That would break literally everything downstream and is a completely unexpected operation...

Issue #71, opened 2026-03-13
Syntalos crashes when attempting a run without selecting a camera in the Aravis module


caenrigen:
[SyntalosCrashReport_261303_2.md](https://github.com/user-attachments/files/25966328/SyntalosCrashReport_261303_2.md)
[SyntalosCrashReport_261303.md](https://github.com/user-attachments/files/25966327/SyntalosCrashReport_261303.md)

Seems reproducible

PR #73, opened 2026-03-13
canvas: Fix black frames for oddly sized frames or crops

caenrigen:
@ximion when cropping (specially via the handy Select Region) you don't always end up with a "nice" frame size, `GL_UNPACK_ALIGNMENT = 4` by default and leads to a black Canvas. 

Very annoying, took a while to get to the bottom of it because I was run into other crashes (#70, #71 , #72) and had no idea what I was really chasing 🥵


ximion:
Oh, that's a nice find, I never ran into this somehow. But this definitely makes sense!

Issue #60, opened 2026-03-04
Mempool [...] has no more space left, MemoryManager: unable to acquire a chunk with a chunk-payload


caenrigen:
@ximion hi again 😝 

Yesterday I run into an issue for which Codex suggested has to do with the control pool memory size. In the shell I got:

```bash
engine: Preparing 'Pupil Labs Neon'...
Mempool [m_chunkSize = 1064, numberOfChunks = 50, used_chunks = 50 ] has no more space left
2026-03-03 17:51:33.610 [ Error ]: MemoryManager: unable to acquire a chunk with a chunk-payload size of 4The following mempools are available:  MemPool [ ChunkSize = 1064, ChunkPayloadSize = 1024, ChunkCount = 50 ]  MemPool [ ChunkSize = 524328, ChunkPayloadSize = 524288, ChunkCount = 50 ]  MemPool [ ChunkSize = 1048616, ChunkPayloadSize = 1048576, ChunkCount = 20 ]  MemPool [ ChunkSize = 6291496, ChunkPayloadSize = 6291456, ChunkCount = 20 ]  MemPool [ ChunkSize = 25165864, ChunkPayloadSize = 25165824, ChunkCount = 10 ]
2026-03-03 17:51:33.611 [Warning]: ICEORYX error! MEPOO__MEMPOOL_GETCHUNK_POOL_IS_RUNNING_OUT_OF_CHUNKS
Error raised by module 'Pupil Labs Neon': Unable to allocate SetNiceness request!
```

<details>
<summary>Click to expand full log</summary>

```bash
Current board settings file: dev.syct
engine: Storing temporary data in: /var/tmp/syntalos-tmprun-tOqMjM
engine: Initializing new ephemeral recording run
engine: Explicit CPU core affinity is disabled.
engine: Running modules in order: Aravis Camera; Canvas; Polar Verity Sense; JSON Writer 2; Python Script; Firmata IO; Video Transformer 2; Canvas 4; Pupil Labs Neon; Canvas 5; Video Transformer; Video Recorder; Polar Verity Sense 2; JSON Writer 3; Video Transformer 3; Video Recorder 3; Shimmer; Plot Time Series; Aravis Camera 2; Canvas 3; Canvas 2; Canvas 6; Video Recorder 2; JSON Writer; 
engine: Obtained system sleep/idle/shutdown inhibitor for this run.
engine: Predicted amount of CPU cores with no (explicitly known) occupation: 17
[...]

PR #79, opened 2026-03-14
[Draft] Fix mem pool issue by refreshing the persistent IPC sender

caenrigen:
@ximion I might have reached the bottom of my rabbit hole 🐇🕳️ 

Codex finally came up with a patch that at least seems to prevent #60 from happening during my testing.

I am not qualified to judge if it really identified the real root cause of #60 or it go lucky with the proposed change.

I used the "Extra High" thinking/effort mode, very curious to know if it was worth spending these LLM tokens 😂

---

Its hypothesis was:

> long-lived iceoryx senders keep one cached chunk each in `m_lastChunkUnmanaged`, and because the worker processes are reused across runs, those cached small control chunks accumulate instead of returning between runs. That matches both the log and the `iceoryx` implementation: `chunk_sender.inl` (line 106), `chunk_sender.inl` (line 193), `chunk_sender.inl` (line 251).

---

If the identified cause is real, perhaps the following Codex suggestions are useful on deciding on the right fix (feel free to ignore, might be slop):

> The clean alternatives, ordered from most pragmatic to most elegant, are:
> 
> **Restart each worker process between runs.** This is the simplest robust fix. It guarantees all iceoryx ports and cached chunks disappear. The downside is extra startup cost and losing any warm in-process state.

I tested this, it works indeed as reported in https://github.com/syntalos/syntalos/issues/60#issuecomment-4056748732

> **Make the worker IPC layer run-scoped instead of process-scoped.** In SyntalosLink, keep module callbacks, timer, and port definitions alive, but create all iceoryx publishers/servers only for one run and destroy them on stop(). This is basically the clean version of the fix we just proved.
> 
> **Refactor the iceoryx ownership into one transport subobject and rebuild that subobject per run.** For example, move pub*, req*, waitset wiring, and output ioxPub ownership into a TransportState owned by SyntalosLink::Private. Then stop() just destroys TransportState and prepare() recreates it. This is cleaner than individually refreshing 20+ members.
> 
> **Make only sender-backed objects run-scoped.** Keep input subscribers and higher-level state persistent, but recreate fixed publishers, fixed servers, and output publishers on each run. This is the smallest architectural change that still targets the real cause.
> 
> **Create low-volume control publishers on demand.** pubSettingsChange, pubInPortChange, pubOutPortChange, and possibly pubStatusMessage do not need to live forever. A temporary publisher per send is simple and avoids keeping cached chunks in those channels.
> 
> **Create output publishers lazily and destroy them on stop.** Output stream publishers are a natural run-scoped resource. This reduces persistent sender count without touching the whole control plane.
> 
> **Collapse many fixed RPC servers into one generic command server.** Right now each worker has many long-lived server objects. A single command server plus a small dispatcher would reduce the number of persistent sender objects a lot. This is elegant, but it is a protocol refactor.
> 
> **Move the control plane off iceoryx entirely.** Keep data streams on iceoryx, but use Qt local sockets, DBus, or another simple IPC mechanism for commands and metadata. This avoids the cached-chunk behavior on the control path. It is the cleanest long-term design, but also the biggest change.
> 
> **Keep the current fix, but formalize it as an explicit reset API.** Instead of “deep refresh” logs and manual member-by-member recreation, add something like SyntalosLink::resetTransportForNextRun(). That keeps the proven behavior while making the code less ad hoc.

### DIFF
--- a/src/mlink/syntaloslink.cpp
+++ b/src/mlink/syntaloslink.cpp
@@ -214,6 +214,7 @@ public:
         reqShutdown = makeServer<ShutdownRequest, DoneResponse>(SHUTDOWN_CALL_ID);
         reqShowDisplay = makeServer<ShowDisplayRequest, DoneResponse>(SHOW_DISPLAY_CALL_ID);
         reqShowSettings = makeUntypedServer(SHOW_SETTINGS_CALL_ID);
+        refreshPersistentIpcSendersPending = false;
     }
 
     ~Private() {}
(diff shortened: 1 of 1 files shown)


caenrigen:
I agree it can do amazing things but for serious code (like Syntalos) it should be used responsibly 😇

Issue #81, opened 2026-03-16
SecondaryClockSynchronizer: potential bug/algorithm change introduced in recent commit


caenrigen:
@ximion  About the `datactl` changes in 05a60feb302c8e7e0440056b4fde7dd4ee8dc916:

In the previous code you had a comment saying:

```cpp
// calculate offsets without the new datapoint included
```

With the latest code it seems that (in Codex' words):

> it inserts first and averages second. That means the new sample partly “votes on whether it is weird.”

Is that true in the latest version? If so, is that intentional? or is this AI slop? 🙈

---

Codex' report:

> `src/datactl/timesync.cpp` (line 630) and `src/datactl/timesync.cpp` (line 638) are not behavior-preserving for `SecondaryClockSynchronizer`. Before this commit, that function explicitly calculated offsets “without the new datapoint included”; now it inserts the current sample before computing `avgOffsetUsec`, so both the tolerance gate and the fluke check at `src/datactl/timesync.cpp` (line 677) judge the sample against a mean already pulled toward itself. That weakens spike rejection and can turn a single moderate timestamp blip into an unnecessary correction/offset notification when the historical average was still within tolerance. The running-sum optimization should preserve the old semantics by computing the average from the pre-insert sum.

PR #82, opened 2026-03-16
datactl: Add moving average regression test

caenrigen:
Codex implemented this to prove #81

### DIFF
--- a/tests/test-timer.cpp
+++ b/tests/test-timer.cpp
@@ -361,6 +361,98 @@ private slots:
         }
     }
 
+    void runExClockSynchronizerMovingAverage()
+    {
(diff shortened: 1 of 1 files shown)


tests/test-timer.cpp
```diff
@@ -5,12 +5,21 @@
 #include <thread>
 
 #include "datactl/syclock.h"
+#define private public
```
ximion: Wow ewwww, that is an insane pattern that I haven't seen before. Not sure if crazy evil, or acceptable for tests. I am very much leaning towards evil though, ideally the function should be testing behavior, not internals. And if it were to look at internals, we may have to think about adding a read-only accessor instead.

PR #102, opened 2026-03-27
[Draft] Rust debug

caenrigen:
Codex modified the build to include the rust debug symbols etc.

Maybe useful for some inspiration

NB fixed small typo, cherry pick it in your commits!


ximion:
TBH, this is already a pretty horrendous hack, and I would rather not like to pile on to it by adding even more options and possible parameters :sweat_smile: 

Ideally I would remove the code completely, but more realistically it will only be used on rare occasions and we can use a packaged iceoryx2 that replaces it in most cases.

The current local build also already includes debug symbols :-)

I would take that typo fix though!

Issue #107, opened 2026-04-01
Feature request: Provide a way to type the `syntalos_mlink` python package when developing Custom Python Modules


caenrigen:
It wound be nice to have autocomplete and typing working when importing `syntalos_mlink`.

Usually you point the IDE (e.g., VS Codium) to the virtualenv and it picks up the python packages and their typings.

My suggestion would be:

Generate and install typing artifacts for `syntalos_mlink` into the venv (always create venv even if no `requirements.txt` is present). Practically that means a `syntalos_mlink.pyi`, or a proper package layout with stubs.

Additional venv invalidation for Syntalos API changes might be a good idea (but this is probably fairly rare).

These necessary files can probably be generated during Syntalos build and then copy-pasted/symlinked into each venv.

Perhpas `pybind11-stubgen syntalos_mlink` could be useful.

caenrigen:
Thanks for the thoughtful reply!

> As for stubs, we can likely create them already - without the other changes, they will just be a fair bit less useful.

Given the complexity required for a more proper solution, I think (auto-)generating a stub file that can be manually added to the IDE settings (or e.g. symlinked into the Python Module `/types` subdirectory) would solve the main pain point of this feature request.

In any case, this is to document the UX, but there is for sure bigger priorities around (e.g. #113)

ximion:
> Given the complexity required for a more proper solution, I think (auto-)generating a stub file that can be manually added to the IDE settings (or e.g. symlinked into the Python Module `/types` subdirectory) would solve the main pain point of this feature request.

Well, the secret knowledge is: If you are on Git master, and have `pybind11-stubgen` installed (available on Debian 13 though the Syntalos repo), then you will already have an interface file that even gets installed alongside the Python module ;-)

But you could also copy it manually if you liked, from the `python` folder after build.

ximion:
@caenrigen Just FYI, since you are using the master branch, Python modules now can (and must!) register their ports dynamically when loaded. There are no more static ports in `module.toml`.

This is part of the ongoing effort to make Python modules a lot less special and make them work like any other OOP module that Syntalos can run. But it *is* an API break (one of multiple that will be needed, unfortunately), so be aware!

caenrigen:
Nice improvement! I will test it this weekend 😄

ximion:
[...]

PR #122, opened 2026-04-20
python: Enhance typing for ports and settings

caenrigen:
Follow up from https://github.com/syntalos/syntalos/issues/107#issuecomment-4282427283

This was mostly Codex-implemented, and with some manual edits by me.

I hope there at least something to learn from it, I doubt that the full PR is usable...

It fixed all the typing issues I was seeing in one of my Python Modules.
The main issues being the ports registering (`Any` typing) and the save/load functions (iirc there was at least a bytes vs bytearray) issue


src/python/pysy-mlink.cpp
```diff
@@ -50,7 +52,10 @@ static SyntalosLink *getActiveLink();
 // Global Python interface object for the Syntalos link
 static PySyLinkManager *g_pslMgr = nullptr;
 
-using PyNewDataFn = std::function<void(const py::object &obj)>;
+using PortData =
+    std::variant<ControlCommand, TableRow, Frame, FirmataControl, FirmataData, IntSignalBlock, FloatSignalBlock>;
+using PyNewDataFn = std::function<void(const PortData &obj)>;
+using OptionalPyNewDataFn = std::optional<PyNewDataFn>;
```
ximion: This is unneeded in two ways: First, the alias is almost as long as the real name, but second, std::function is itself already nullable, and pybind11 knows this - so you can just drop it ;-)

src/python/pysy-mlink.cpp
```diff
@@ -281,7 +278,16 @@ struct OutputPort {
     const std::shared_ptr<OutputPortInfo> _oport;
 };
 
-using PySaveSettingsFn = std::function<ByteVector(const fs::path &baseDir)>;
```
ximion, line 287: NO! `ByteVector` is a known type for pybind11, it will transparently convert it to bytes/bytearray, so using py::bytes explicitly just incurs additional overhead when writing the code, as well as possibly more copies of the data too.

src/python/pysy-mlink.cpp
```diff
@@ -281,7 +278,16 @@ struct OutputPort {
...
+using PySaveSettingsFn = std::function<py::bytes(const fs::path &baseDir)>;
+using PyLoadSettingsFn = std::function<bool(py::bytes, const fs::path &baseDir)>;
+using OptionalPrepareRunFn = std::optional<PrepareRunFn>;
+using OptionalStartFn = std::optional<StartFn>;
+using OptionalStopFn = std::optional<StopFn>;
+using OptionalShowSettingsFn = std::optional<ShowSettingsFn>;
+using OptionalShowDisplayFn = std::optional<ShowDisplayFn>;
+using OptionalPySaveSettingsFn = std::optional<PySaveSettingsFn>;
+using OptionalPyLoadSettingsFn = std::optional<PyLoadSettingsFn>;
+using OptionalEventFn = std::optional<std::function<void()>>;
```
ximion: Do not add these aliases, and do not use std::optional on std::function in general - it just adds extra wrappers for no reason at all.

src/python/pysy-mlink.cpp
```diff
@@ -402,14 +408,16 @@ class PySyLinkManager
         });
     }
 
-    auto onPrepare()
+    OptionalPrepareRunFn onPrepare() const
     {
+        if (!m_prepareFn)
+            return std::nullopt;
```
ximion: This, and the other changes like it, are nonsense changes, as changing absolutely nothing on that function would have had the exact same effect... - this just adds more code to achieve the exact same thing... Weird.

src/python/pysy-mlink.cpp
```diff
@@ -454,14 +464,16 @@ class PySyLinkManager
...
-    auto onStop()
+    OptionalStopFn onStop() const
     {
+        if (!m_stopFn)
+            return std::nullopt;
         return m_stopFn;
     }
 
-    void setOnShowSettings(ShowSettingsFn fn)
+    void setOnShowSettings(const OptionalShowSettingsFn &fn)
```
ximion: Passing by const reference makes us loose the advantage of moving the parameter, so we copy the reference twice, and also potentially loose track of it on the Python side (depending on whether it is ultimately implicitly moved or copied) - don't do that!

src/python/pysy-mlink.cpp
```diff
@@ -567,14 +588,16 @@ class PySyLinkManager
      * @param data_type   Data type ID (e.g. DataType.Frame, DataType.TableRow).
      * @returns InputPort handle, or None if registration failed (e.g. duplicate ID).
      */
-    py::object registerInputPort(const std::string &id, const std::string &title, BaseDataType::TypeId data_type)
+    std::optional<InputPort> registerInputPort(
```
ximion, line 574: This on the other hand, is alright - but we may want to throw an exception in case of just returning nothing, I'll have to look into that...

ximion (changes requested):
Lots of changes that do nothing or make things worse this time 😅 
The AI didn't understand how the bindings work in some places...

One thing I am less sure about is the std::variant bit - currently, we convert the type to a py::object and that back into the type. If a std::variant is in the mix, we go type -> variant-wrapped -> pyobj and back again. Given how the types work, I am worried that this might mean that we now copy the data an additional time, but may that's not how it works. This one I will have to look into.

src/python/pysy-mlink.cpp
caenrigen: raise exception sounds better

caenrigen:
> Lots of changes that do nothing or make things worse this time 😅 The AI didn't understand how the bindings work in some places...

Sorry for the AI slop, Matthias! 🙈🙈🙈 Was a combination of not great prompt, rebasing my codebase mid conversation, running out of Codex usage, and of course me not know anything about python binding 🤦🏻‍♂️

Just wanted to share a quick draft that was able to fix the missing types.

I did some cleanups, but there is still perhaps too many changes... Feel free to cherrypick anything that might be useful in your own commits and close this PR 🙈

ximion:
I find the visitor pattern over a std::variant quite elegant, and would have liked to keep it - but looking into the copy/move semantics for variants and BaseDataType led me down a completely different rabbot hole - I'm exploring that first, then get back to this one :-)

ximion:
I made my own version of this PR, which makes some light API behavior changes and is slightly more flexible.

The `std::variant` pattern in the original PR, before you removed it, was very interesting actually - the code looked really clean. If this pattern works without copying data (expensive!) and without performance impact from the lambdas (probably there, but also probably so tine that it can be ignored) then this might be something we may want to do, simply because the code is really clean, and it might allow us to define all datatypes in only one place, instead of (currently) touching the Python module and datactl both (realistically though, one will always have to touch Python as well to add converters for new data types...).

caenrigen:
I am glad the PR contributed with some new insights 🙂

ximion:
If you want to bring that part back, make a new PR :-)

Otherwise I might play with it some later time ^^
The original issue should be addressed now, at least :-D

PR #123, opened 2026-04-21
python: Add dummy Firmata control output port to surface bug

caenrigen:
@ximion Firmata Python Module interface got broken in the process of the migration, this is more of an issue than a PR, but up to you if you want to keep another smoke test around

```
py.arduino_led(LED Controller): Event loop tick failed: RuntimeError(return_value_policy = copy, but type Syntalos::FirmataControl is non-copyable!)
Traceback (most recent call last):
  File "/home/ilab/.local/share/Syntalos/modules/arduino_led/mod-main.py", line 148, in event_loop_tick
    _ = ctl.firmata_register_digital_pin(
        STATE.settings.pin_start, "START_PULSE_PIN", is_output
    )
RuntimeError: return_value_policy = copy, but type Syntalos::FirmataControl is non-copyable!
```

The PR just adds a dummy port to the Python Script to surface the bug (Codex-implemented, did not unpack it to compare the changes, etc.)

I suppose you will know what the underlying problem is.


ximion:
Ah, nice, this is exactly the type of performance issue that my change was trying to catch - it's just rather unfortunate that this happens at runtime, that's a bit bizarre. In any case, I need to decide what needs to be copied and what can be passed by reference (we probably need to copy almost everything, as it's safer to hand over full control to Python).

ximion:
It's actually visible in the file diff (easy to create with the syct-to-text debugging tool) that Syntalos didn't write this file :-P

Using `register_*` functions in PyScript is also something that you shouldn't do. Still, this is a good thing to add as a test, I am merging this change and update the test later.

This issue is also already fixed in master (the CI agrees! ^^)

PR #125, opened 2026-04-22
python: Use callback-specific payload wrappers to fix callback types

caenrigen:
@ximion I am starting to feel the "careful what you wish for" 😂😂😂

Automatically generating the stub file is very neat and is the type of thing that appeals to me a lot, but it is not a magic bullet sadly. Since we are already in the rabbit hole... we might as well finish the job

Did you spot that the "wrong side" of the `io_name(side1, side2)` is used for the callback functions? The stub generation process has no way of knowing that these are callback functions that Python "provides" and C++ feeds.

In other words, changing to `io_name` did not really solve the issue for callback functions because the automated stub generation doesn't understand the real direction of the data flow for callback functions. 🤷🏼‍♂️ 

---

This PR fixes both the `bytearray | bytes` vs `bytes` issue and the `os.PathLike | str | bytes` vs `pathlib.Path`.

This time I had a more careful ping pong with Codex.

There seems to be no better solution while generating the stub file automatically.

It is a relatively large amount of "manual" code for a pesky issue... 😅

---

PS in case you want to make the casting more explicit you might want to also apply this:

```cpp
return m_loadSettingsFn(PyCallbackBytes {ByteVector(settings)}, baseDir);
```

but I kept it to minimal changes

## Summary by CodeRabbit

* **API Changes**
  * Settings callback `on_save_settings` now strictly requires returning `bytes` type
  * Settings callback `on_load_settings` parameter type narrowed to `bytes` (not bytearray)
  * Path parameters standardized to use `pathlib.Path` for consistency

### DIFF
--- a/src/python/pysy-mlink.cpp
+++ b/src/python/pysy-mlink.cpp
@@ -50,6 +50,104 @@ static PySyLinkManager *g_pslMgr = nullptr;
 
 using PyNewDataFn = std::function<void(const py::object &obj)>;
 
+// Restrict Python callback signatures to pathlib.Path while keeping fs::path internally.
+struct PyCallbackPath {
+    PyCallbackPath() = default;
+    PyCallbackPath(fs::path p)
+        : path(std::move(p))
+    {
+    }
+
+    fs::path path;
+};
+
+// Use callback-specific payload wrappers so nested Callable[...] stubs reflect
+// what Python code actually receives and returns.
+struct PyCallbackBytes {
+    PyCallbackBytes() = default;
+    PyCallbackBytes(ByteVector bytes)
+        : data(std::move(bytes))
+    {
+    }
+
+    ByteVector data;
+};
+
+struct PyCallbackBytesLike {
+    PyCallbackBytesLike() = default;
+    PyCallbackBytesLike(ByteVector bytes)
+        : data(std::move(bytes))
+    {
+    }
+
+    ByteVector data;
+};
+
+namespace pybind11::detail
+{
+
+template<>
+struct type_caster<PyCallbackPath> {
+    PYBIND11_TYPE_CASTER(PyCallbackPath, io_name("pathlib.Path", "pathlib.Path"));
+
+    bool load(handle src, bool convert)
+    {
+        make_caster<fs::path> path_caster;
+        if (!path_caster.load(src, convert))
+            return false;
+
+        value = PyCallbackPath{cast_op<fs::path>(path_caster)};
+        return true;
+    }
+
+    static handle cast(const PyCallbackPath &src, return_value_policy policy, handle parent)
+    {
+        return py::cast(src.path, policy, parent).release();
+    }
+};
+
+template<>
+struct type_caster<PyCallbackBytes> {
+    PYBIND11_TYPE_CASTER(PyCallbackBytes, io_name("bytes", "bytes"));
+
+    bool load(handle src, bool convert)
+    {
+        make_caster<ByteVector> bytes_caster;
+        if (!bytes_caster.load(src, convert))
+            return false;
+
+        value = PyCallbackBytes {cast_op<ByteVector>(bytes_caster)};
+        return true;
+    }
+
+    static handle cast(const PyCallbackBytes &src, return_value_policy policy, handle parent)
+    {
+        return py::cast(src.data, policy, parent).release();
+    }
+};
+
+template<>
+struct type_caster<PyCallbackBytesLike> {
+    PYBIND11_TYPE_CASTER(PyCallbackBytesLike, io_name("bytes | bytearray", "bytes | bytearray"));
+
+    bool load(handle src, bool convert)
+    {
+        make_caster<ByteVector> bytes_caster;
+        if (!bytes_caster.load(src, convert))
+            return false;
+
+        value = PyCallbackBytesLike {cast_op<ByteVector>(bytes_caster)};
+        return true;
+    }
+
+    static handle cast(const PyCallbackBytesLike &src, return_value_policy policy, handle parent)
+    {
+        return py::cast(src.data, policy, parent).release();
+    }
+};
+
+} // namespace pybind11::detail
+
 SyntalosPyError::SyntalosPyError(const char *what_arg)
     : std::runtime_error(what_arg) {};
 SyntalosPyError::SyntalosPyError(const std::string &what_arg)
@@ -319,7 +417,8 @@ struct OutputPort {
     const std::shared_ptr<OutputPortInfo> _oport;
 };
 
-using PySaveSettingsFn = std::function<ByteVector(const fs::path &baseDir)>;
+using PySaveSettingsFn = std::function<PyCallbackBytesLike(const PyCallbackPath &baseDir)>;
+using PyLoadSettingsFn = std::function<bool(const PyCallbackBytes &settings, const PyCallbackPath &baseDir)>;
 
 /**
  * Central link object for Syntalos Python modules.
@@ -339,7 +438,7 @@ private:
     StopFn m_stopFn;
 
     PySaveSettingsFn m_saveSettingsFn;
-    LoadSettingsFn m_loadSettingsFn;
+    PyLoadSettingsFn m_loadSettingsFn;
 
     ShowSettingsFn m_showSettingsFn;
     ShowDisplayFn m_showDisplayFn;
@@ -571,7 +670,8 @@ public:
 
         m_slink->setSaveSettingsCallback([this](ByteVector &settings, const fs::path &baseDir) {
             try {
-                settings = m_saveSettingsFn(baseDir);
+                auto pySettings = m_saveSettingsFn(baseDir);
+                settings = std::move(pySettings.data);
                 return true;
             } catch (py::error_already_set &e) {
                 if (!handlePyError(m_slink, e))
@@ -586,7 +686,7 @@ public:
         return m_saveSettingsFn;
     }
 
-    void setOnLoadSettings(LoadSettingsFn fn)
+    void setOnLoadSettings(PyLoadSettingsFn fn)
     {
         m_loadSettingsFn = std::move(fn);
         if (!m_loadSettingsFn) {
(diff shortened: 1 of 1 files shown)


ximion:
> I am starting to feel the "careful what you wish for" 😂😂😂

It's just about as bad as I expected it to be, tbh - and of course, I am piling on to it with the API changes that come as a consequence of the Qt-free Syntalos bindings. The good news on that front is that since today, those changes are complete! So there shouldn't be significant API changes anymore (famous last words, but at least all the major work is done now... I didn't expect to actually do this for the next release, it was an insane amount of work).

On this PR:

I need to think about this some more - I see what the PR does and it makes sense to me, however, the generated stub data without this PR looks mostly correct to me already:
```python
@property
def on_save_settings(self) -> collections.abc.Callable[[os.PathLike | str | bytes], bytes]:
    ...
@on_save_settings.setter
def on_save_settings(self, arg1: collections.abc.Callable[[os.PathLike | str | bytes], bytes]) -> None:
    ...
```

The only thing that's "wrong" is the callable return type, which could be `bytes | bytearray`, but the stub generator thinks that only `bytes` is permitted.

In any case, if we add this workaround, I'd likely want it in a separate file - the main one is already very cluttered, and I want to avoid having automatic pybind11 typecasters in there.

src/python/pysy-mlink.cpp
```diff
@@ -50,6 +50,104 @@ static PySyLinkManager *g_pslMgr = nullptr;
 
 using PyNewDataFn = std::function<void(const py::object &obj)>;
 
+// Restrict Python callback signatures to pathlib.Path while keeping fs::path internally.
+struct PyCallbackPath {
```
ximion, line 54: What is the purpose of this? I think you could just drop this, pybind11 handles `std::filesystem` types well on its own already.

caenrigen:
Good morning, Matthias!

> It's just about as bad as I expected it to be, tbh - and of course, I am piling on to it with the API changes that come as a consequence of the Qt-free Syntalos bindings. The good news on that front is that since today, those changes are complete! So there shouldn't be significant API changes anymore (famous last words, but at least all the major work is done now...

🥳🥳🥳

The changes in the Python API are great! 😄😄😄 It makes a lot more sense, and way less "hidden magic". The way it is now possible to pump an event loop is much cleaner and neat. 🙏🏻🙏🏻🙏🏻

> I didn't expect to actually do this for the next release, it was an insane amount of work).

🫂❤️ if you are around Netherlands (Rotterdam area) dinner is on me! 😉 

> I need to think about this some more - I see what the PR does and it makes sense to me, however, the generated stub data without this PR looks mostly correct to me already:
> 
> ```python
> @property
> def on_save_settings(self) -> collections.abc.Callable[[os.PathLike | str | bytes], bytes]:
>     ...
> @on_save_settings.setter
> def on_save_settings(self, arg1: collections.abc.Callable[[os.PathLike | str | bytes], bytes]) -> None:
>     ...
> ```
> 
> The only thing that's "wrong" is the callable return type, which could be `bytes | bytearray`, but the stub generator thinks that only `bytes` is permitted.

I am not sure I communicated the issue properly.

Both `os.PathLike | str | bytes` and `bytes` are actually wrong regarding to what the current code is "communicating" via the usage of `io_name`.

For `on_save_settings` it should be:

```python
collections.abc.Callable[[pathlib.Path], bytes | bytearray]
```

Under the hood:

The wrong side of the `io_name` arguments is being picked up from your:

https://github.com/syntalos/syntalos/blob/b2faf8220b0081381ace0e43b08d5a9469c75ff5/src/python/sydatatopy.h#L229

And for the file paths:

https://github.com/pybind/pybind11/blob/678b6735e44fe9d7feac7ff31bc9c86d42c04613/include/pybind11/stl/filesystem.h#L99

i.e. it can accept any of the `os.PathLike | str | bytes` but it always outputs `pathlib.Path`. Which makes sense to me, "very Pythonic".

The process of generating the stub file is not smart enough (or simply lacks the information) to understand that "C++ code will be calling these Python callbacks passing in data from the C++ side into their arguments".

Let me know if you understand what I mean by the "wrong side of `io_name(textSide1, textSide2)`". I might not have the correct words or concepts to explain it properly. This is my understanding of what is going wrong from my ping pong with Codex.

> In any case, if we add this workaround, I'd likely want it in a separate file - the main one is already very cluttered, and I want to avoid having automatic pybind11 typecasters in there.

Perfectly fine by me 🙂

src/python/pysy-mlink.cpp
caenrigen: My reply below might explain this too (yes pybind11 knows how to handle it, but it does not understand the direction of the data flow, it assumes the wrong direction)

ximion:
674097096_10163500088795528_5250443691631547179_n

ximion:
Thanks!

> The way it is now possible to pump an event loop is much cleaner and neat.

Pumping an event loop sucks though, but the lifecycle guarantees of the FDs associated with iox aren't entirely clear to me yet, so I don't want to depend on that just yet. Chances are also that we may have multi-threaded MLink modules in future, which would also change the game a little (currently, modules are single-threaded and blocking, so, if their data consumers are too slow, the upstream modules will get blocked, unlike library modules which just continue submitting data uninterrupted until their queue gets too full and they get stopped).

> 🫂❤️ if you are around Netherlands (Rotterdam area) dinner is on me! 😉

Thank you! I actually will be in the Netherlands soon, but near Nijmegen - a bit far from Rotterdam :-)

> For on_save_settings it should be

You may be looking at it from a "what does this actually submit" perspective, in which it is true that it'll *always* submit a pathlib.Path right now to the callback. I am looking at it more from an API contract side, in that, with the type being `os.PathLike[str]` I am saying "be ready to accept anything path-like". Which is what I do want to say here, the receiving function *should* be ready to handle anything that looks like a path, even though we are currently only sending one thing.

I do understand the `ByteVector` case though, where the pybind11 typing assumes that since the C++ type resolves to `bytes` as an output, the function can only return `bytes` and not also a `bytearray`. It is a bit strange to me that this is the case though, because for a return type, pybind11 should figure this out - it already does so in other cases that aren't callbacks.

Plus, in C++, data flow direction can usually very reliably be inferred. I haven't had time to look at this in-depth yet, from my quick look a while ago though I couldn't find specific API for it in pybind11. Which probably makes some sense, type annotations and automatic interface generation are extremely new for pybind11, this stuff hasn't been battle-tested yet.

caenrigen:
> Thank you! I actually will be in the Netherlands soon, but near Nijmegen - a bit far from Rotterdam :-)

let's see, I pinged you on LinkedIn ;-)

caenrigen:
> You may be looking at it from a "what does this actually submit" perspective, in which it is true that it'll _always_ submit a pathlib.Path right now to the callback.

To great extent I was looking from that perspective, but also from these perspectives:

- "keep it simple": if you "force" the callback signatures to accept all the possible path-like objects then my modules code has to be more elaborate (with effectively dead code) for the sake of covering all the possible path-like objects. I am putting aside the fact that so far I am not using this path at all.
- "I just wanted the type checker to work and not complain": e.g. iirc when using `os.PathLike` (on the C++ side) it does not become `os.PathLikep[str]` in the stub file and the type checker will either complain that `os.PathLike != os.PathLike[str]` or it will complain that `os.PathLike` is missing `[str]` 😂. Not 100% if this was the issue I was running into but it lead to a situation where I could not run away from typing errors due to other idiosyncrasies.

> I am looking at it more from an API contract side, in that, with the type being `os.PathLike[str]` I am saying "be ready to accept anything path-like". Which is what I do want to say here, the receiving function _should_ be ready to handle anything that looks like a path, even though we are currently only sending one thing.

I agree that in general functions that receive path-like arguments should be ready to accept anything that ducks like a ~duck~ path, it makes it very convenient to use such functions without extra boiler plate. However, in the case of Syntalos' callbacks, I don't think the callback functions will ever be used/called by anything else besides Syntalos itself, on the Python side it is a specialized narrow-scoped function. There is more to gain from simplicity than a broad API contract.

> It is a bit strange to me that this is the case though, because for a return type, pybind11 should figure this out - it already does so in other cases that aren't callbacks.

iirc from my chat with Codex, the problem is that the types end up inside a `Callable[ARG, RETURN]` and being placed as an `ARG` or as a `RETURN` leads to a binding that assumes the wrong data flow direction (Python to C++ vs vice-versa). The main use case of Python bindings is to provide Python with an interface to call C++ functions (not the other way around, which is Syntalos' callbacks use case). I suppose that is the underlying "bug"

caenrigen:
I heard good stuff about CodeRabbit 😄 let's see!

ximion:
It was recommended to me by fellow maintainers to reduce the load of reviewing so much stuff. And if it's not as terrible as Copilot is, and manual review is done in addition, then it at least won't hurt...

You are the guinea pig with this MR here, because it's reasonably complex and on a less public project ;-)

And also because I wonder if it would come up with something better somehow than this :sweat_smile:

ximion:
> And also because I wonder if it would come up with something better somehow than this 😅

Sadly it didn't, but wrote a poem about this PR instead... :laughing: 

I definitely wouldn't want the custom path conversions, I think those are definitely wrong.

> iirc when using os.PathLike (on the C++ side) it does not become os.PathLikep[str] in the stub file and the type checker will either complain that os.PathLike != os.PathLike[str] or it will complain that os.PathLike is missing [str]

Using `os.PathLike | str` isn't wrong here though... I wonder if you are hitting https://github.com/pybind/pybind11-stubgen/issues/266 and that's what the type checker trips over? I don't want to work around this and rather address it at the source.

I also wonder if we can override the return type with a docstring instead...

ximion:
The rabbit did not impress. But, to be fair to it, your commit is sane from a C++ perspective, and its summary and highlights are very accurate :-)

I really, really don't like this workaround though, so I addressed this upstream in https://github.com/pybind/pybind11/pull/6055

Let's see if that is a change that is viable to them, or if there is any better solution.

caenrigen:
> then it at least won't hurt...

there is a price to pay: additional cognitive load to read and process the (hopefully not) slop

> Sadly it didn't, but wrote a poem about this PR instead... 😆

I though you were joking 😂😂😂 what a waste of tokens (see if you can activate [caveman mode](https://github.com/JuliusBrussee/caveman))

> The rabbit did not impress. But, to be fair to it, your commit is sane from a C++ perspective, and its summary and highlights are very accurate :-)

This is not the best PR for Rabbit to shine its capabilities, this PR is asking the Rabbit "do you like my hack? 🥕" 

---

> I wonder if you are hitting https://github.com/pybind/pybind11-stubgen/issues/266

The stub file I got generated is missing the `import os` indeed, but my type checker trips in some other unexpected ways when my Python functions do not match 100% the typing in the stub file, I will restrain myself from elaborating on that for now

> Using `os.PathLike | str` isn't wrong here though...

Type checker complains (inside the the stub file itself) with `Expected type arguments for generic class "PathLike"`, i.e., it wants something like `os.PathLike[str]` instead, no idea why `os.PathLike` is even a generic...

> I really, really don't like this workaround though, so I addressed this upstream in https://github.com/pybind/pybind11/pull/6055

Fixing this upstream is the proper way, thanks for working to it! 💪🏼

ximion:
FWIW, the change is merged upstream: https://github.com/pybind/pybind11/commit/5a5f21deb6c1dbfe8fc36547419d188ce7ab30cc - This means you'll have a proper solution at zero cost and complexity as soon as a new pybind11 version is released. Or you can test the current Git snapshot of pybind11 ;-)

caenrigen:
I noticed indeed (I was subscribed to your PR upstream, well done 🙌🏼)

Feel free to close my PR

ximion:
I'd leave a bug report open (for reference) until a new upstream release of the 3rd-party component has been made, but you're right, this PR can probably be closed.

Thank you so much for your patience and for raising the issue and proposing a solution!

PR #126, opened 2026-04-22
python: Enhance example-py typing

caenrigen:
Follow up from #125

### DIFF
--- a/modules/example-py/mod-main.py
+++ b/modules/example-py/mod-main.py
@@ -15,7 +15,7 @@ module from scratch. If you want to do that though, this
 example is a good starting point.
 """
 
-import os
+import pathlib
 import sys
 import syntalos_mlink as syl
 
@@ -26,7 +26,6 @@ from PyQt6.QtWidgets import (
     QLineEdit,
     QPushButton,
     QVBoxLayout,
-    QHBoxLayout,
 )
 import json
 import cv2 as cv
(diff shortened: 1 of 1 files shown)


ximion:
On the one hand, that's a bit too many type hints for my taste (anything in function bodies should be inferred in most cases, for simple things). On the other hand, this is an example, and the type hints are very helpful for new users, even as simple documentation pointers.

So yeah, let's have this :-)

modules/example-py/mod-main.py
```diff
@@ -172,22 +173,22 @@ def on_dialog_closed():
...
+    def _save_settings_data(self, baseDir: pathlib.Path) -> bytes:
         """
         Serialize settings to bytes so Syntalos can store them for us.
         """
 
         settings = dict(prefix=self._label_prefix)
         return bytes(json.dumps(settings), 'utf-8')
 
-    def _load_settings_data(self, data: bytes, baseDir: os.PathLike[str]) -> bool:
+    def _load_settings_data(self, data: bytes, baseDir: pathlib.Path) -> bool:
```
ximion, line 184: This is a strong specialization, that I am not extremely happy with - I'd like *some* flexibility when returning types, as locking yourself in to exactly one type isn't great in Python when many compatible types could also be emitted. That said, in this case, PathLike and Path are pretty much the same thing. And we don't return a string here ever, since the API is locked into returning a path.

caenrigen:
> So yeah, let's have this :-)

I filled in a lot of typing inside the methods  only because the basedpyright typechecker requires "@final" decorator in order to allow implicit typing.

But I agree that more typing helps people that are getting started

PR #129, opened 2026-04-24
python: Allow using print() instead of syl.println()

caenrigen:
See comment in the code.

Is there some deeper reason why `syl.println()` exists in the Python module, compared to what this PR proposes?

Maybe it is needed for PyScript, though that one has its own logging window under the editor 🤔

In any case it would be nice to support Python's native `print()`, it is way more natural to use inside Python. Beside that some dependencies packages might be printing relevant things too that get "suppressed" (i.e. buffered) until closing Syntalos, at which point all the prints are flushed.

This PR should also allow to use a proper logger (e.g. Python's `logging`)

### DIFF
--- a/src/moduleloader-py.cpp
+++ b/src/moduleloader-py.cpp
@@ -123,10 +123,14 @@ public:
         if (!existingPythonPath.isEmpty())
             pythonPath << existingPythonPath.split(':', Qt::SkipEmptyParts);
         pythonPath.removeDuplicates();
-        if (!pythonPath.isEmpty()) {
+        if (!pythonPath.isEmpty())
             penv.insert(QStringLiteral("PYTHONPATH"), pythonPath.join(':'));
-            setModuleBinaryEnv(penv);
-        }
+
+        // Python stdout is block-buffered when connected to QProcess pipes, which
+        // makes ordinary print() output appear only when the worker exits. Keep
+        // module output interactive so print() behaves like syl.println().
+        penv.insert(QStringLiteral("PYTHONUNBUFFERED"), QStringLiteral("1"));
+        setModuleBinaryEnv(penv);
 
         // Resolve interpreter: use the venv's python when a venv is requested.
         // setPythonVirtualEnv() tells runProcess() to set VIRTUAL_ENV and prepend


ximion:
The only reason `println` exists is because `PYTHONUNBUFFERED` never worked even once during my testing. It was always set in `pyworker`, but had no effect.

Does this work now? (in any case, we can set it for Python modules, since we are already setting it for pyworker, just to keep the two identical, but, I don't have high hopes given past experiences...

caenrigen:
I did a quick check and it seemed to work for me, but I will verify this in some more detail

caenrigen:
Seems to work fine:

```
18:15:45.663047 syntalos:mod.pupil_labs_neon-0: 18:15:45.662825: prepare()
18:15:45.951076 syntalos:mod.pupil_labs_neon-0: 18:15:45.946610: prepare() done
18:15:45.953793 syntalos:engine: Module 'Pupil Labs Neon' prepared in 313 msec
...
18:15:45.965893 syntalos:engine: Exporting streams for external modules...
18:15:45.968433 syntalos:engine: Initializing launch...
18:15:45.969092 syntalos:engine: Started event thread 'm:plot-timeseries' with 2 participating modules
18:15:45.969093 syntalos:engine: Module and engine threads created in 0 msec
18:15:45.969094 syntalos:engine: Waiting for 'Pupil Labs Neon' to get ready...
18:15:46.226379 syntalos:engine: Waited for modules to get ready for 257 msec
18:15:46.226380 syntalos:engine: Launch setup completed.
18:15:46.230846 syntalos:engine: Started system resource monitoring.
18:15:46.281373 syntalos:mod.pupil_labs_neon-0: 18:15:46.231609: start()
18:15:46.281373 syntalos:mod.pupil_labs_neon-0: 18:15:46.231695: start() done
```

ximion:
Oh, awesome! I last tried this ages ago (maybe with Python 3.10) and nothing got it to work. `syl.println` was the workaround because I needed that as a debugging tool and wasted a ton of time on unbuffered Python already, and had experiments to run ^^

Glad it works now, maybe the fact that the Python modules are now standalone has something to do with it...

PR #131, opened 2026-04-24
mlink: Allow show settings to be invoked when module is in ERROR state

caenrigen:
This is the solution that has been preventing my Python Modules to open the settings after an Exception has been raised, either handled (via `syntax_mlink.raise()`) or unhandled (letting the exception be raised in Python, which is handy since it now prints the traceback in red in `syntalos` output.

Currently the function passed to `await_data_forever()` continues to be called in the ERROR state, so I would say there is no reason to not allow show settings to be called too


ximion:
This is a safe change, can you also apply it to `showDisplayUi()`?

PR #132, opened 2026-04-25
python: Fix setproctitle fallback

caenrigen:
Follow up from edaca57b0250f8e5be0f4758611be8120d645f4d

The fallback logic was faulty, newOptn not used.

Codex explanation (before this PR):

> init_link(rename_process=True) now hard-requires setproctitle, despite the fallback path.
pysy-mlink.cpp (line 335) directly imports setproctitle; if it is not installed, pybind throws before initSyntalosModuleLink() can do the simpler /proc/self/comm fallback. Also, newOptn is computed but not used at pysy-mlink.cpp (line 343), so even when setproctitle exists, the intended “prefer setproctitle over simple rename” path is not actually applied.

### DIFF
--- a/src/python/pysy-mlink.cpp
+++ b/src/python/pysy-mlink.cpp
@@ -332,19 +332,24 @@ public:
             // argv[], which Python does not provide. Rather than doing the horrendous
             // memory-scanning of setproctitle ourselves, we might as well call the real
             // thing, if it is installed.
-            sptMod = py::module_::import("setproctitle");
+            try {
+                sptMod = py::module_::import("setproctitle");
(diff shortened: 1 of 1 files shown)


ximion:
It doesn't fix what it thinks it fixes, but managed to fix a real bug by accident - I did  not consider that the import might throw, so, adding that exception handling is good!

caenrigen:
😂 upsss 

The import did throw indeed on my first attempt so I sent it to investigate for issues

ximion:
Yeah, I should have tested the "what if setproctitle isn't installed?" scenario as well, instead of assuming that this API was exception-free.

PR #133, opened 2026-04-25
Firmata: Preserve qFatal() behavior and abort on critical errors

caenrigen:
Codex explanation:

> Critical Firmata validation errors are no longer fatal.
Several qFatal() calls were changed to LOG_CRITICAL() in backend.cpp (line 245), backend.cpp (line 257), backend.cpp (line 269), etc. Quill critical logging does not abort, so invalid pin/channel paths now return normally instead of stopping on a caller contract violation. Two messages also still use %d, so the pin value will not be formatted correctly with Quill/fmt.

I assume the behaviour change was not intentional, hence the PR

### DIFF
--- a/modules/firmata-io/firmata/backend.cpp
+++ b/modules/firmata-io/firmata/backend.cpp
@@ -254,7 +254,7 @@ void FirmataBackend::writeDigitalPin(uint8_t pin, bool value)
         writeBuffer(buffer, sizeof(buffer));
 
     } else {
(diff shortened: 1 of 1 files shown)


ximion:
No, that was 100% intentional. The previous behavior was stupid, we should not crash the entire application just because the user made a mistake and supplied a bad pin value. Now, we just log the error and can continue without a crash, since receiving a bad value doesn't end up in any inconsistent state whatsoever.

So, previous behavior was wrong, new behavior is much less crazy :-)

caenrigen:
Roger that 🫡

PR #134, opened 2026-04-25
trivial: More logging cleanup that passed by unnoticed on QDebug removal

caenrigen:
Codex explanation:

> Build regression: QDebug was removed while stream-style Qt logging remains.
The range removes <QDebug> from several files, but they still use qCritical()/qWarning() streaming syntax, which requires the QDebug type. Examples: stream.h (line 674), tracker.cpp (line 44), tomlutils.cpp (line 133), sysinfo.cpp (line 159). This can fail compilation depending on transitive includes.

seems that `<QDebug>` is still sneakily included

### DIFF
--- a/modules/triled-tracker/tracker.cpp
+++ b/modules/triled-tracker/tracker.cpp
@@ -27,6 +27,8 @@
 #include <opencv2/highgui/highgui.hpp>
 #include <opencv2/imgproc/imgproc.hpp>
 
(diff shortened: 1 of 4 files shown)


ximion:
It all ends up with Quill eventually, through the log handlers ;-)

Using the direct logging functions means less indirections and more metadata though, so this is a good change! Thanks!

PR #135, opened 2026-04-25
mlink: Send stop request when module is in ERROR state

caenrigen:
A change similar to 056e335851ce66617b1f83a761758da492725501.

Why: If my Python Module raises an exception during `prepare()` and afterwards I try to save the settings it fails:

```bash
17:15:54.809839 syntalos:board.save: Saving board as: /home/victor/utm/emc/emc_ilab_daq/boards/arduino_led.syct
17:15:54.816597 syntalos:main: Current board settings file: /home/victor/utm/emc/emc_ilab_daq/boards/arduino_led.syct
17:15:56.759218 syntalos:engine: Storing temporary data in: /var/tmp/syntalos-tmprun-xKrJUb
17:15:56.759220 syntalos:engine: Initializing new ephemeral recording run
17:15:56.775160 syntalos:engine: Explicit CPU core affinity is disabled.
17:15:56.775255 syntalos:engine: Module start order: LED Controller
17:15:56.775255 syntalos:engine: Module stop order:  LED Controller
17:15:56.785184 syntalos:engine: Obtained system sleep/idle/shutdown inhibitor for this run.
17:15:56.785189 syntalos:engine: Predicted amount of CPU cores with no (explicitly known) occupation: 6
17:15:56.785307 syntalos:engine: Preparing 'LED Controller'...
17:15:56.800828 syntalos:mod.arduino_led-0: Error raised by 'LED Controller': RuntimeError: Some dummy error
17:15:56.800828 syntalos:mod.arduino_led-0: 
17:15:56.800828 syntalos:mod.arduino_led-0: At:
17:15:56.800828 syntalos:mod.arduino_led-0:   /home/victor/.local/share/Syntalos/modules/arduino_led/mod-main.py(104): prepare
17:15:56.800828 syntalos:mod.arduino_led-0:   /home/victor/.local/share/Syntalos/modules/arduino_led/mod-main.py(268): main
17:15:56.800828 syntalos:mod.arduino_led-0:   /home/victor/.local/share/Syntalos/modules/arduino_led/mod-main.py(273): <module>
17:15:59.665032 syntalos:engine: Module 'LED Controller' failed to prepare.
17:15:59.668430 syntalos:engine: Run stopped, finalizing...
17:15:59.669472 syntalos:engine: Waited 0 msec for event threads to stop.
17:15:59.669473 syntalos:engine: Stopping 'LED Controller'...
17:15:59.670493 syntalos:engine: Module 'LED Controller' stopped in 0 msec
17:15:59.670494 syntalos:engine: Stopping IPC stream exporter...
17:15:59.670964 syntalos:engine: Joining remaining threads.
17:15:59.671226 syntalos:engine: All (non-event) engine threads joined in 0 msec
17:15:59.671227 syntalos:engine: Removing broken data...
17:15:59.681118 syntalos:engine: Ready.
17:15:59.681255 syntalos:engine: Removing temporary storage directory
17:15:59.681320 syntalos:engine: Ephemeral run completed (result: success)
17:16:04.829994 syntalos:board.save: Saving board as: /home/victor/utm/emc/emc_ilab_daq/boards/arduino_led.syct
17:16:04.832673 syntalos:mod.arduino_led-0: Failed to save settings (issue communicating with the module). Reusing old settings.
17:16:04.832964 syntalos:mod.arduino_led-0: 0 [W] Receiver { connections: PolymorphicVec<iceoryx2_bb_concurrency::cell::UnsafeCell<core::option::Option<iceoryx2_bb_container::slotmap::SlotMapKey>>, iceoryx2_bb_memory::heap_allocator::HeapAllocator> { capacity: 2, len: 2 content: [ UnsafeCell(UnsafeCell { .. }), UnsafeCell(UnsafeCell { .. }) ] }, receiver_port_id: 3104346086804211321435731933221, service_state: ServiceState { dynamic_storage: Storage { shm: SharedMemory { name: FileName { value: StaticString<255> { len: 92, data: "sy_305ad9523c6b202364d581359ec3d2c5743e42e7_cb35bc73827b8835ad85a8e8aa4b6cd1d15cfec1.dynamic" } }, has_ownership: AtomicBool(false), memory_mapping: MemoryMapping { file_descriptor: Some(FileDescriptor { value: 99, is_owned: true }), file_path: None, base_address: 0xffff9c5ef000, size: 747, offset: 0 }, memory_lock: None, mapping_offset: 0 }, name: FileName { value: StaticString<255> { len: 40, data: "cb35bc73827b8835ad85a8e8aa4b6cd1d15cfec1" } }, _phantom_data: PhantomData<iceoryx2::service::dynamic_config::DynamicConfig> }, additional_resource: NoResource, static_config: StaticConfig { service_hash: ServiceHash(RestrictedFileName { value: StaticString<64> { len: 40, data: "cb35bc73827b8835ad85a8e8aa4b6cd1d15cfec1" } }), service_name: ServiceName { value: StaticString<255> { len: 29, data: "Sy/arduino_led_0/SaveSettings" } }, unique_service_id: UniqueServiceId(UniqueSystemId { value: 3025116134162561838846170511397, pid: 669733, creation_time: Time { clock_type: Monotonic, seconds: 52245, nanoseconds: 783116986 } }), attributes: AttributeSet(StaticVec<iceoryx2::service::attribute::Attribute, 8> { len: 0, content: [  ] }), messaging_pattern: RequestResponse(StaticConfig { enable_safe_overflow_for_requests: true, enable_safe_overflow_for_responses: true, enable_fire_and_forget_requests: true, max_active_requests_per_client: 4, max_loaned_requests: 2, max_response_buffer_size: 2, max_servers: 2, max_clients: 2, max_nodes: 4, max_borrowed_responses_per_pending_response: 2, request_message_type_details: MessageTypeDetails { header: TypeDetail { variant: FixedSize, type_name: StaticString<256> { len: 58, data: "iceoryx2::service::header::request_response::RequestHeader" }, size: 56, alignment: 8 }, user_header: TypeDetail { variant: FixedSize, type_name: StaticString<256> { len: 2, data: "()" }, size: 0, alignment: 1 }, payload: TypeDetail { variant: Dynamic, type_name: StaticString<256> { len: 19, data: "__cxx__abi__St4byte" }, size: 1, alignment: 1 } }, response_message_type_details: MessageTypeDetails { header: TypeDetail { variant: FixedSize, type_name: StaticString<256> { len: 59, data: "iceoryx2::service::header::request_response::ResponseHeader" }, size: 48, alignment: 8 }, user_header: TypeDetail { variant: FixedSize, type_name: StaticString<256> { len: 2, data: "()" }, size: 0, alignment: 1 }, payload: TypeDetail { variant: Dynamic, type_name: StaticString<256> { len: 19, data: "__cxx__abi__St4byte" }, size: 1, alignment: 1 } } }) }, shared_node: SharedNode { id: UniqueNodeId(UniqueSystemId { value: 13922977192892559638781376549, pid: 669733, creation_time: Time { clock_type: Monotonic, seconds: 52245, nanoseconds: 754766106 } }), details: NodeDetails { executable: FileName { value: StaticString<255> { len: 10, data: "python3.13" } }, name: NodeName { value: StaticString<128> { len: 13, data: "arduino_led_0" } }, config: Config { global: Global { root_path: Path { value: StaticString<255> { len: 27, data: "/run/user/1000/syntalos-iox" } }, prefix: FileName { value: StaticString<255> { len: 3, data: "sy_" } }, service: Service { directory: Path { value: StaticString<255> { len: 8, data: "services" } }, data_segment_suffix: FileName { value: StaticString<255> { len: 5, data: ".data" } }, static_config_storage_suffix: FileName { value: StaticString<255> { len: 8, data: ".service" } }, dynamic_config_storage_suffix: FileName { value: StaticString<255> { len: 8, data: ".dynamic" } }, creation_timeout: 500ms, connection_suffix: FileName { value: StaticString<255> { len: 11, data: ".connection" } }, event_connection_suffix: FileName { value: StaticString<255> { len: 6, data: ".event" } }, blackboard_mgmt_suffix: FileName { value: StaticString<255> { len: 16, data: ".blackboard_mgmt" } }, blackboard_data_suffix: FileName { value: StaticString<255> { len: 16, data: ".blackboard_data" } } }, node: Node { directory: Path { value: StaticString<255> { len: 5, data: "nodes" } }, monitor_suffix: FileName { value: StaticString<255> { len: 13, data: ".node_monitor" } }, static_config_suffix: FileName { value: StaticString<255> { len: 8, data: ".details" } }, service_tag_suffix: FileName { value: StaticString<255> { len: 12, data: ".service_tag" } }, cleanup_dead_nodes_on_creation: true, cleanup_dead_nodes_on_destruction: true } }, defaults: Defaults { publish_subscribe: PublishSubscribe { max_subscribers: 8, max_publishers: 2, max_nodes: 20, subscriber_max_buffer_size: 2, subscriber_max_borrowed_samples: 2, publisher_max_loaned_samples: 2, publisher_history_size: 0, enable_safe_overflow: true, unable_to_deliver_strategy: Block, subscriber_expired_connection_buffer: 128 }, event: Event { max_listeners: 16, max_notifiers: 16, max_nodes: 36, event_id_max_value: 255, deadline: None, notifier_created_event: None, notifier_dropped_event: None, notifier_dead_event: None }, request_response: RequestResonse { enable_safe_overflow_for_requests: true, enable_safe_overflow_for_responses: true, max_active_requests_per_client: 4, max_response_buffer_size: 2, max_servers: 2, max_clients: 8, max_nodes: 20, max_borrowed_responses_per_pending_response: 2, max_loaned_requests: 2, server_max_loaned_responses_per_request: 2, client_unable_to_deliver_strategy: Block, server_unable_to_deliver_strategy: Block, client_expired_connection_buffer: 128, enable_fire_and_forget_requests: true, server_expired_connection_buffer: 128 }, blackboard: Blackboard { max_readers: 8, max_nodes: 20 } } } }, monitoring_token: UnsafeCell(UnsafeCell { .. }), registered_services: RegisteredServices { handle: MutexHandle { handle: HandleStorage<libc::unix::linux_like::linux::pthread_mutex_t> { is_interprocess_capable: false, is_initialized: true }, clock_type: UnsafeCell(UnsafeCell { .. }), value: UnsafeCell(UnsafeCell { .. }) } }, signal_handling_mode: HandleTerminationRequests, _details_storage: Storage { name: FileName { value: StaticString<255> { len: 4, data: "node" } }, config: Configuration { path: Path { value: StaticString<255> { len: 63, data: "/run/user/1000/syntalos-iox/nodes/13922977192892559638781376549" } }, suffix: FileName { value: StaticString<255> { len: 8, data: ".details" } }, prefix: FileName { value: StaticString<255> { len: 3, data: "sy_" } } }, has_ownership: AtomicBool(false), file: File { path: Some(FilePath { value: StaticString<255> { len: 79, data: "/run/user/1000/syntalos-iox/nodes/13922977192892559638781376549/sy_node.details" } }), file_descriptor: FileDescriptor { value: 18, is_owned: true }, access_mode: ReadWrite, has_ownership: AtomicBool(false) }, len: 1777 } }, static_storage: Storage { name: FileName { value: StaticString<255> { len: 40, data: "cb35bc73827b8835ad85a8e8aa4b6cd1d15cfec1" } }, config: Configuration { path: Path { value: StaticString<255> { len: 36, data: "/run/user/1000/syntalos-iox/services" } }, suffix: FileName { value: StaticString<255> { len: 8, data: ".service" } }, prefix: FileName { value: StaticString<255> { len: 3, data: "sy_" } } }, has_ownership: AtomicBool(false), file: File { path: Some(FilePath { value: StaticString<255> { len: 88, data: "/run/user/1000/syntalos-iox/services/sy_cb35bc73827b8835ad85a8e8aa4b6cd1d15cfec1.service" } }), file_descriptor: FileDescriptor { value: 98, is_owned: true }, access_mode: ReadWrite, has_ownership: AtomicBool(false) }, len: 1519 } }, buffer_size: 4, tagger: CyclicTagger(4), to_be_removed_connections: Some(UnsafeCell(UnsafeCell { .. })), degradation_callback: None, message_type_details: MessageTypeDetails { header: TypeDetail { variant: FixedSize, type_name: StaticString<256> { len: 58, data: "iceoryx2::service::header::request_response::RequestHeader" }, size: 56, alignment: 8 }, user_header: TypeDetail { variant: FixedSize, type_name: StaticString<256> { len: 2, data: "()" }, size: 0, alignment: 1 }, payload: TypeDetail { variant: Dynamic, type_name: StaticString<256> { len: 19, data: "__cxx__abi__St4byte" }, size: 1, alignment: 1 } }, receiver_max_borrowed_samples: 4, enable_safe_overflow: true, number_of_channels: 1, connection_storage: UnsafeCell(UnsafeCell { .. }), initial_channel_state: ChannelState(0) } 
17:16:04.832964 syntalos:mod.arduino_led-0: | Lost a sample. This only happens in the dynamic use case when a sender has reallocated its data segment and gone out of scope before the receiver has mapped the realloacted data segment. To circumvent this, you could either use static memory or increase the initial max slice len.
```

idk if this is the right solution... Codex was suggesting more crazy fixes (that did not work) while this one was a side note.

With this patch the module returns back to `Idle` instead of remaining in `Error` state. It kind of makes sense, it is ready to try again etc.. But not sure what the intention is. Before this patch the module remained red 🤷🏼‍♂️ 

---

btw, is there supposed to be a difference between how syntalos treats Python modules sending a `syl.raise_error(msg)` vs an actual (unhandled) `Exception` being raised in Python? The latter is more handy because the traceback gets nicely printed in red in `syntalos` output

### DIFF
--- a/src/fabric/mlinkmodule.cpp
+++ b/src/fabric/mlinkmodule.cpp
@@ -1310,7 +1310,7 @@ void MLinkModule::runThread(OptionalWaitCondition *startWaitCondition)
 void MLinkModule::stop()
 {
     if (isProcessRunning())
-        d->callClientSimple<StopRequest>(this, STOP_CALL_ID, [](auto &) {}, 15);
+        d->callClientSimple<StopRequest>(this, STOP_CALL_ID, [](auto &) {}, 15, true, false);
 
     // stop the module thread first
     AbstractModule::stop();


ximion:
> With this patch the module returns back to `Idle` instead of remaining in `Error` state. It kind of makes sense, it is ready to try again etc.. But not sure what the intention is. Before this patch the module remained red 🤷🏼‍♂️

That is the intended behavior. If something crashed, the user should be able to see what died and - ideally - why, even after dismissing the error message. The error state is cleared when it is explicitly dismissed by trying to run the experiment again.

The commit message is wrong, but this patch is still correct: The stop request is *sent*, but we do not wait for a reply... We really should wait at that point though.

> btw, is there supposed to be a difference between how syntalos treats Python modules sending a `syl.raise_error(msg)` vs an actual (unhandled) `Exception` being raised in Python? The latter is more handy because the traceback gets nicely printed in red in `syntalos` output

If an unhandled exception is raised, the process dies and the communication channel vanishes, so the user only gets a "something crashed for some reason" message. If you actually send an error message with `raise_error`, it gets displayed to the user and the process continues to run, leading to a recoverable state.

The former is a crash, the latter is a handled error state.

caenrigen:
> That is the intended behavior. If something crashed, the user should be able to see what died and - ideally - why, even after dismissing the error message. The error state is cleared when it is explicitly dismissed by trying to run the experiment again.

The error message remains, even though the state returns to `Idle`, so in principle it is still possible to identify which module was in `Error` state. No strong opinions on this

> If an unhandled exception is raised, the process dies and the communication channel vanishes, so the user only gets a "something crashed for some reason" message.

Conceptually I understand and it makes sense, but for python (modules) are you sure that is the case?

It seems to me that Python is running in some kind of interactive mode that allows Syntalos to invoke the callbacks. If that is the case, then there is no real process exit when typical python `Exception`s are raised (of course python could still crash at some deeper level due to bugs in the python interpreter itself, or killed OOM, etc.). Instead it behaves more like a notebook, the `Exception` is raised but you can run code again afterwards without restarting the python process.

From a quick check it did not seem that the Python process is restarted after an exception.

ximion:
> The error message remains, even though the state returns to `Idle`, so in principle it is still possible to identify which module was in `Error` state. No strong opinions on this

The message might, but if the status is clear, visual indicators like the red box and red LED vanish. Those are quite handy to immediately visually pinpoint the culprit :-)

> Conceptually I understand and it makes sense, but for python (modules) are you sure that is the case?

Yes

> It seems to me that Python is running in some kind of interactive mode that allows Syntalos to invoke the callbacks.

Not anymore. That was one of the reasons for the `pyworker` process, but modules aren't using that anymore. This is a feature reserved only for `pyscript` now (and even there, getting a proper message and recovering the interpreter from error states was rough).

If you raise an exception now, your Python process will die just like any other process. It's no different from an unhandled C++ exception or Rust panic in that regard.

ximion:
> Conceptually I understand and it makes sense, but for python (modules) are you sure that is the case?

I actually briefly forgot that for callbacks, the C++ module has some nice logic which does do the retracing and communication, and prevents the module from dying - **if** the exception was raised from a callback (and almost anything is callback-based...).

While playing with it, I did find out though that I accidentally made Python modules almost immortal, they couldn't crash in callbacks even if they really needed to. That's been fixed now though. You still get the benefit of having exceptions forwarded and the process not die, but if the Python process needs to die, we now let it.

ximion:
"Immortal Python" is certainly one of the best accidental rock band names ever...

caenrigen:
I was crafting the message below when I saw your new comments  😂😂😂

---

I am not sure I understand correctly.

As of 8755df40be0a5b2a12b9d08edd878933171d7720, is the python process supposed to be terminated and restarted when there is some unhandled `Exception` raised in the Python Module?

Is saving the Python Module settings (via board save) after an unhandled `Exception` supposed to work (without opening the settings GUI)? It does not work as of 8755df40be0a5b2a12b9d08edd878933171d7720. Using `syl.raise_error("Bla")` has the same warning that settings could be saved for the module, opening the settings GUI solves it (it clears the error state). Anyway, this is fairly rare edge case.

caenrigen:
> "Immortal Python" is certainly one of the best accidental rock band names ever...

😂 had no idea that was a band!

> You still get the benefit of having exceptions forwarded and the process not die

imo, given that the process is not killed (and indeed there is no need to kill it), then this warning should be fixed after Python raised an exception and we press the button to save the board:

```bash
21:24:36.825770 syntalos:mod.arduino_led-0: Failed to save settings (issue communicating with the module). Reusing old settings.
```

This PR fixed initially, but it is happening again (goes away only after I open the settings GUI, I guess because it clears the `Error` state back to `Idle` and the rest of the logic associated with that transition)

ximion:
> 😂 had no idea that was a band!

Not yet, but it could be... If I start one, this would be a great name ^^

> Is saving the Python Module settings (via board save) after an unhandled Exception supposed to work (without opening the settings GUI)?

Just to make sure we are on the same page, "unhandled exception" means that nothing handled it and the Python process crashed (as opposed to the other case, where the exception may not be handled in Python, but the callback exception handler handles it, and thus doesn't cause a crash).

Restarting the process just to serialize would be bad, because we may accidentally override the project data with bad data from the just-restarted process. However, if you get that warning and the process did **not** die, then something else is wrong, because then communication should still work.

I'll have a look.

ximion:
> I'll have a look.

There was indeed a bug. Things should work as intended again now :-)

caenrigen:
Sorry, I meant exceptions in the callbacks (e.g. in prepare()) that my python code did not handle

ximion:
Yeah, that's an unhandled exception by you, but in the grand scheme of things, a handled exception in Syntalos.

NGL, I am at this point quite happy about the `syntalos_mlink` Python module - it does some pretty smart things! However of course, that only matters if the master doesn't do buggy stuff, but that's resolved now and I finally understand the bug you were trying to fix the entire time.

Sorry that it took so long to grasp that!

caenrigen:
No worries, thank you for the pacience and quick patches! 🚀🚀🚀

ximion:
It was a busy week for me, with a major Linux security vulnerability to fix - this stuff is almost relaxing :laughing: 

And, of course, the project benefits greatly from your testing, because any issue you find now is one that we don't release with and that users who do not have the expertise to test the master branch won't hit.

By the way, unrelated permissions question: Are you able to download the build artifacts on pages like this one? => https://github.com/syntalos/syntalos/actions/runs/24890930254

caenrigen:
> It was a busy week for me, with a major Linux security vulnerability to fix

😧

is that the [Pack2TheRoot](https://github.security.telekom.com/2026/04/pack2theroot-linux-local-privilege-escalation.html)?

The `xz` story of a while ago makes one question how many critical systems out there are already compromised 😅

> this stuff is almost relaxing 😆

a refreshing breeze in the grand ocean of bugs 😂

> And, of course, the project benefits greatly from your testing, because any issue you find now is one that we don't release with and that users who do not have the expertise to test the master branch won't hit.

🫡 😇 

> By the way, unrelated permissions question: Are you able to download the build artifacts on pages like this one? => https://github.com/syntalos/syntalos/actions/runs/24890930254

Yes, I can download the files 🙂

caenrigen:
It was Pack2TheRoot indeed hehe 💪🏼 

```
2026-04-13: First draft of private patch by PackageKit maintainer Matthias Klumpp ([@ximion](https://github.com/ximion))
```

ximion:
> The `xz` story of a while ago makes one question how many critical systems out there are already compromised 😅

The impact of that incident on the maintainer community is still quite extreme... Passing on a project to someone that you haven't met at a conference yet became a lot harder.

I could definitely tell stories...

Pack2TheRoot sucked, but it wasn't unexpected - I was pondering at a conference last January why there hasn't been such an issue yet, given how great of a target `packagekitd` is for security researchers (it's a daemon that, by design, cannot be sandboxed - lots of opportunities there).

We are legit currently drowning in AI-generated security issues and other bugs, and unlike a year ago, this time the percentage of slop is significantly lower.

> Yes, I can download the files 🙂

Excellent, so all you need is a GitHub account then to access those. For people without one, I created https://github.com/syntalos/syntalos/releases/tag/snapshot now as well where the system will make snapshot builds available automatically. For the brave ;-) (and for debugging purposes).

caenrigen:
> The impact of that incident on the maintainer community is still quite extreme...

was also a good wake up call to take security and dependencies even more seriously

> Passing on a project to someone that you haven't met at a conference yet became a lot harder.

trust is a tricky business 😅

> I could definitely tell stories...

Would be glad to hear them some day 😉 

> We are legit currently drowning in AI-generated security issues and other bugs

I heard about this too, bug bounties are also flooded with slop

> and unlike a year ago, this time the percentage of slop is significantly lower.

better models? 🤔

ximion:
> better models? 🤔

In part yes, but also models in the hand of people who know exactly where to direct them to :sweat_smile:

caenrigen:
> In part yes, but also models in the hand of people who know exactly where to direct them to 😅

Sadly the capability to exploit bugs is becoming increasingly democratized for bad actors too, it is always a cat and mouse game but I am not sure it is really balanced right now

> makes one question how many critical systems out there are already compromised 😅

btw, just heard about CVE-2026-31431 (aka "Copy Fail"), and got email from an HPC having to do emergency reboot due to critical vulnerability... 😅 "fun" times

ximion:
We certainly need better funding, across nearly all of open source...

TBH, "Copy fail" was handled pretty badly in my opinion, because patches weren't ready to be deployed in downstream distributions and a working exploit was made available immediately too. Would have been nice to have waited a bit after informing distros, to give them a head start, and also wait at least a few days with publishing the exploit... But oh well, one vulnerability less is always a good outcome :-)

caenrigen:
> We certainly need better funding, across nearly all of open source...

Not a direct money funding, but perhaps you could quality for [Project Glasswing](https://www.anthropic.com/project/glasswing) to help you harden the security of your open source projects 🙂 

> TBH, "Copy fail" was handled pretty badly in my opinion, because patches weren't ready to be deployed in downstream distributions and a working exploit was made available immediately too. Would have been nice to have waited a bit after informing distros, to give them a head start, and also wait at least a few days with publishing the exploit... But oh well, one vulnerability less is always a good outcome :-)

A tricky thing to know how much time is enough time, and if the good or the bad guys get an edge from hearing about it earlier than later. Some systems might have been already compromised earlier in which case I suppose it is better for everyone to hear about the vulnerability earlier than later.

ximion:
> Not a direct money funding, but perhaps you could quality for [Project Glasswing](https://www.anthropic.com/project/glasswing) to help you harden the security of your open source projects 🙂

Haha, I ndon't think they consider my projects or the Linux desktop important enough. But access to something like this would go through a large corporation or organization anyway. It's far more likely that people from those will just come with their bug reports.

> A tricky thing to know how much time is enough time, and if the good or the bad guys get an edge from hearing about it earlier than later.

A published exploit while patches aren't distributed yet helps mostly the attackers...

PR #139, opened 2026-04-26
mlink: Bypass master queues for direct MLink streams

### LINKED ISSUE #138: Memory leak again 😭
No idea if it is a real leak but a board like this (python module feeding frames to another python module) leads to Syntalos process memory exploding and eventually crashing/getting killed by the OS:

Image

Smells like some of the issue I had before, was there some regression maybe?

caenrigen:
Codex-implemented

> Added direct-IPC capability API via usesDirectModuleLink() in moduleapi.h (line 513), defaulting false, with MLinkModule overriding true in mlinkmodule.cpp (line 648).
>
> Marked MLink-to-MLink graph subscriptions as direct IPC bypass at connection time in moduleapi.cpp (line 274).
>
> Added bypass handling to stream subscriptions so bypassed subscriptions clear pending data and do not enqueue/clone incoming frames in stream.h (line 389).
>
> Added hasDataSubscribers() / dataSubscriberCount() so Syntalos can distinguish real in-process consumers from direct IPC consumers in stream.h (line 763).
> 
> Changed MLink output forwarding so the master process does not subscribe to a worker output when all downstream consumers are direct MLink modules in mlinkmodule.cpp (line 1085).
> 
> Updated StreamExporter direct-IPC fast path to use the new capability API and mark the subscription bypassed in streamexporter.cpp (line 135).

Seems to fix #138

### DIFF
--- a/src/fabric/mlinkmodule.cpp
+++ b/src/fabric/mlinkmodule.cpp
@@ -645,6 +645,11 @@ ModuleDriverKind MLinkModule::driver() const
     return ModuleDriverKind::THREAD_DEDICATED;
 }
 
+bool MLinkModule::usesDirectModuleLink() const
+{
+    return true;
+}
+
 ModuleFeatures MLinkModule::features() const
 {
     return ModuleFeature::SHOW_DISPLAY | ModuleFeature::SHOW_SETTINGS;
@@ -1077,7 +1082,7 @@ bool MLinkModule::registerOutPortForwarders()
 
     // connect to external process streams
     for (auto &oport : outPorts()) {
-        if (!oport->streamVar()->hasSubscribers())
+        if (!oport->streamVar()->hasDataSubscribers())
             continue;
 
         Private::OutPortSub ps;
--- a/src/fabric/mlinkmodule.h
+++ b/src/fabric/mlinkmodule.h
@@ -59,6 +59,7 @@ public:
     bool initialize() override;
 
     ModuleDriverKind driver() const override;
+    bool usesDirectModuleLink() const override;
     virtual ModuleFeatures features() const override;
 
     QString moduleBinary() const;
--- a/src/fabric/moduleapi.cpp
+++ b/src/fabric/moduleapi.cpp
@@ -276,6 +276,12 @@ void VarStreamInputPort::setSubscription(StreamOutputPort *src, std::shared_ptr<
     d->outPort = src;
     m_sub = sub;
 
+    if (sub != nullptr) {
+        const bool useDirectIpc =
+            src != nullptr && src->owner()->usesDirectModuleLink() && d->owner->usesDirectModuleLink();
+        sub->setDirectIpcBypass(useDirectIpc);
+    }
+
     d->owner->inputPortConnected(this);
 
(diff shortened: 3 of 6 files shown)


ximion:
I think I might already be working on something similar, but with a bit more maintainable API (and just as an efficiency improvement, I wasn't aware of the bug before).

I'll have to have a look later though to see if it's really the same thing.

caenrigen:
No worries, I will not need Syntalos till Tuesday, and thos patch seems to work for the purpose of developing my python modules

ximion:
I implemented a different solution as part of an efficiency optimization that I was already working on :-)

In this part of the code, you unfortunately notice that MLink modules weren't in Syntalos' initial design, and were added later... That part of the code can probably be slightly redesigned in future to be more clear, but at least even now it's pretty efficient (just really complex).

caenrigen:
> I implemented a different solution

Codex might have spotted an edge case (a bit elaborate for me to grasp if it a real issue):

> Pure MLink-to-MLink outputs can miss start() metadata updates. In mlinkmodule.cpp (line 1091), allMLink skips creating an OutPortSub; later start() (line 1272) only calls startStream() for d->outPortSubs. That means pure direct MLink outputs do not get the second metadata snapshot after worker start(). This conflicts with the documented intent for modules that disable async start to publish metadata in start() (CONTRIBUTING.md (line 363)). Downstream MLink modules then compare stale input metadata at mlinkmodule.cpp (line 1241) and may skip UpdateInputPortMetadataRequest.

ximion:
That is niche, but probably real... It's easily fixable, but at this point I am starting to wonder whether this design is still viable - instead of having MLink bolted on, the ports API may need a better design that has it included from the ground up.

caenrigen:
> That is niche, but probably real... It's easily fixable, but at this point I am starting to wonder whether this design is still viable - instead of having MLink bolted on, the ports API may need a better design that has it included from the ground up.

Let me know if I can help.

I do not know what exactly you mean but I could send Codex to vibe code some prototypes if you want to explore some potential alternatives 🙂

ximion:
Thanks! But I did a review of this design, and it's overall actually way better than I thought, and was well done in the past - I was just overcomplicating it.

Any new design would for sure not be better, and either be a worse abstraction or leak IPC tasks into the general-purpose SPSC-based streams, which is also not what I want. The current compartmentalized design is actually pretty good, and the fix for that (rather academic) issue that Codex raised was trivial to do. I just simplified things along the way a little.

Potentially, with the entire delayed-metadata-change on start() stuff could even be entirely removed... It solely exists to break cycles in the module connection graph, where the output metadata created by module A depends on the metadata on its input that is coming from module B, but B's metadata itself depends on its input that it receives from A. In that case, the only way correct metadata can be obtained is if one module waits a bit and doesn't need its data in prepare().

This case is now extremely rare though, it once existed with the Firmata modules, but the new engine algorithms to smartly flatten the cyclic graph into a DAG-like structure likely solve this issue for most real-world cases. We may still need it though in rare events, just in case...

caenrigen:
Nice, thanks Matthias! 💪🏼 🚀

PR #155, opened 2026-05-19
Aravis: Restore image transform and propagate rotated output to size metadata

caenrigen:
## Summary by CodeRabbit

* **Bug Fixes**
  * Camera output size now correctly reflects rotation-aware width/height adjustments so reported image dimensions match rotated ROI.
  * Restored camera settings now trigger recalculation of image transform so ROI and pixel-format updates are applied consistently after loading preferences.

[Review Change Stack](https://app.coderabbit.ai/change-stack/syntalos/syntalos/pull/155?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/modules/camera-arv/araviscameramodule.cpp
+++ b/modules/camera-arv/araviscameramodule.cpp
@@ -119,8 +119,9 @@ public:
         m_tfParams = m_configWindow->currentTransformParams();
 
         const auto roi = m_camera->getROI();
-        m_expectedWidth = roi.width();
-        m_expectedHeight = roi.height();
+        const bool swapsAxes = m_tfParams->rot == 1 || m_tfParams->rot == 3;
+        m_expectedWidth = swapsAxes ? roi.height() : roi.width();
+        m_expectedHeight = swapsAxes ? roi.width() : roi.height();
 
         // set the required stream metadata for video capture
         m_outStream->setMetadataValue("size", MetaSize(m_expectedWidth, m_expectedHeight));
--- a/modules/camera-arv/configwindow.cpp
+++ b/modules/camera-arv/configwindow.cpp
@@ -888,6 +888,7 @@ void ArvConfigWindow::loadSettings(const QVariantHash &settings, const QByteArra
             }
         }
     }
+    updateImageTransform();
 
     // ensure any timers run to update the list of available cameras or modify settings
     QApplication::processEvents();


modules/camera-arv/araviscameramodule.cpp
```diff
@@ -121,9 +121,12 @@ class AravisCameraModule : public AbstractModule
         const auto roi = m_camera->getROI();
         m_expectedWidth = roi.width();
         m_expectedHeight = roi.height();
+        const bool swapsAxes = m_tfParams->rot == 1 || m_tfParams->rot == 3;
+        const int width = swapsAxes ? m_expectedHeight : m_expectedWidth;
```
ximion:
```suggestion
        m_expectedWidth = swapsAxes ? roi.height() : roi.width();
```

modules/camera-arv/araviscameramodule.cpp
```diff
@@ -121,9 +121,12 @@ class AravisCameraModule : public AbstractModule
         const auto roi = m_camera->getROI();
         m_expectedWidth = roi.width();
         m_expectedHeight = roi.height();
```
ximion: Drop those two assignments, then:

modules/camera-arv/araviscameramodule.cpp
```diff
@@ -121,9 +121,12 @@ class AravisCameraModule : public AbstractModule
...
+        const bool swapsAxes = m_tfParams->rot == 1 || m_tfParams->rot == 3;
+        const int width = swapsAxes ? m_expectedHeight : m_expectedWidth;
+        const int height = swapsAxes ? m_expectedWidth : m_expectedHeight;
```
ximion:
```suggestion
        m_expectedHeight = swapsAxes ? roi.width() : roi.height();
```

ximion (changes requested):
It is weird to me that the ROI dimensions would not be accurate... Kinda looks like a bug there, the rotation should be taken into account for the ROI... But if this is needed for a small patch without lots of changes to the QArv code, then the patch is okay. It's just strange that the ROI would be wrong to begin with.

modules/camera-arv/araviscameramodule.cpp
caenrigen: My initial solution was kind of correct...

> 122-124: ⚠️ Potential issue | 🔴 Critical | ⚡ Quick win
> 
> Frame validation will fail when rotation swaps axes.
> 
> The decoded image at line 227 has dimensions roi.width() × roi.height() (raw camera output), but the validation check at line 230 runs before software transforms are applied. When rot is 1 or 3, m_expectedWidth and m_expectedHeight are set to the swapped dimensions, causing the validation to reject valid frames with an "unexpected dimensions" error.
> 
> The output stream metadata at line 127 should reflect the post-transform dimensions (swapped), but m_expectedWidth/m_expectedHeight must match the pre-transform decoded frames for validation to work.

I will wait for you decide what solution is best. Feel free to close the PR and push your own version to speed this up 😵‍💫

caenrigen:
> It is weird to me that the ROI dimensions would not be accurate... Kinda looks like a bug there, the rotation should be taken into account for the ROI...

NB the point is to solve these for 90/270 deg rotations:

```
[libvpx-vp9 @ 0x7f00a7c30400] v1.15.0
terminate called after throwing an instance of 'std::runtime_error'
  what():  Received bigger frame than we expected for Video Recorder (480x640 instead of 640x480)
Aborted (core dumped)
```

and same thing different module (line breaks are mine):

```
12:09:39.099110 E ev:m:videotrans:mod.videotransform-0:
Error raised by 'Video Transformer':
Received frame with unexpected dimensions 480x640 (expected 640x480).
The video source is sending invalid frames.
```

ximion:
Thank you for finding this issue! And you are right, my suggestion breaks things elsewhere - this is because `m_expected*` are overloaded values in meaning and try to serve two contracts at once (input/output) while they can't due to the applied software rotations.

I'll merge this and resolve the issue in a separate patch.

PR #157, opened 2026-05-20
camera-generic: Bug fixes and warnings on failures

caenrigen:
This is mostly Codex implemented. Main issue was that the gain was not being set and setting settings would pass by silently (made them emit warnings).

Maybe these improvements are useful to others.

Feel free to close this if it is too many changes. I played a bit with one of the camera and it seemed to work ok.

## Summary by CodeRabbit

* **New Features**
  * Centralized camera property handling with backend-verified application and tolerance-based mismatch warnings
  * Framerate now supports and reports fractional (floating-point) values end-to-end

* **Bug Fixes**
  * Fail-fast behavior and clearer diagnostics when camera connection fails
  * Warnings when device-reported properties (resolution, pixel format, tuning) differ from requested values

* **Improvements**
  * UI and settings persist FPS as a floating-point value for finer configuration

[Review Change Stack](https://app.coderabbit.ai/change-stack/syntalos/syntalos/pull/157?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/modules/camera-generic/camera.cpp
+++ b/modules/camera-generic/camera.cpp
@@ -19,7 +19,9 @@
 
 #include "camera.h"
 
+#include <array>
 #include <QFileInfo>
+#include <cmath>
 #include <fcntl.h>
 #include <libv4l2.h>
 #include <linux/videodev2.h>
@@ -27,6 +29,46 @@
 #include <opencv2/videoio.hpp>
 #include <sys/ioctl.h>
 #include "datactl/frametype.h"
+
+struct CameraPropertyInfo
+{
+    int id;
+    const char *name;
+};
+
+static QString cameraPropertyName(int propertyId)
+{
+    static const std::array<CameraPropertyInfo, 11> propertyNames = {{
+        {cv::CAP_PROP_FRAME_WIDTH, "frame width"},
+        {cv::CAP_PROP_FRAME_HEIGHT, "frame height"},
+        {cv::CAP_PROP_FPS, "framerate"},
+        {cv::CAP_PROP_EXPOSURE, "exposure"},
+        {cv::CAP_PROP_BRIGHTNESS, "brightness"},
+        {cv::CAP_PROP_CONTRAST, "contrast"},
+        {cv::CAP_PROP_SATURATION, "saturation"},
+        {cv::CAP_PROP_HUE, "hue"},
+        {cv::CAP_PROP_GAIN, "gain"},
+        {cv::CAP_PROP_AUTO_EXPOSURE, "auto exposure"},
+        {cv::CAP_PROP_FOURCC, "pixel format"},
+    }};
+
+    for (const auto &property : propertyNames) {
+        if (property.id == propertyId)
+            return QString::fromUtf8(property.name);
+    }
+
+    return QStringLiteral("OpenCV property %1").arg(propertyId);
+}
+
+static bool isPositiveFinite(double value)
+{
+    return std::isfinite(value) && value > 0.0;
+}
+
+static bool differsFromRequested(double reported, double requested, double tolerance = 0.5)
+{
+    return !std::isfinite(reported) || std::abs(reported - requested) >= tolerance;
+}
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wpadded"
@@ -46,7 +88,9 @@ public:
     cv::VideoCapture *cam{};
     int camId;
 
-    int fps{};
+    // Fractional rates exist, e.g. V4L2: `Interval: Discrete 0.133s (7.500 fps)`
+    double fps{};
+    double activeFps{};
     cv::Size frameSize;
     CameraPixelFormat captureFormat;
 
@@ -77,6 +121,7 @@ Camera::Camera(QuillLogger *logger)
     // set some default values
     d->frameSize = cv::Size(960, 720);
     d->fps = 30;
+    d->activeFps = d->fps;
     d->exposure = 10;
     d->brightness = 0;
     d->contrast = 0;
@@ -111,6 +156,56 @@ void Camera::fail(const QString &msg)
     d->lastError = msg;
 }
 
+std::expected<double, QString> Camera::setCameraProperty(int propertyId, double value, bool check, double tolerance)
+{
+    if (!d->cam || !d->cam->isOpened())
+        return std::unexpected(QStringLiteral("Camera is not open."));
+
+    const auto name = cameraPropertyName(propertyId);
+    if (!d->cam->set(propertyId, value)) {
+        const auto cameraValue = d->cam->get(propertyId);
+        const auto msg = QStringLiteral(
+                             "Failed to set camera property '%1' to %2. Camera reports %3. "
+                             "For more OpenCV video I/O details, run Syntalos with: "
+                             "OPENCV_LOG_LEVEL=DEBUG OPENCV_VIDEOIO_DEBUG=1")
+                             .arg(name)
+                             .arg(value)
+                             .arg(cameraValue);
+        LOG_WARNING(
+            d->log,
+            "{}",
+            msg);
+        return std::unexpected(msg);
+    }
+
+    const auto reportedValue = d->cam->get(propertyId);
+    if (check)
+        warnIfCameraReportsDifferent(propertyId, value, reportedValue, tolerance);
+
+    return reportedValue;
+}
+
+void Camera::warnIfCameraReportsDifferent(
+    int propertyId,
+    double requestedValue,
+    double reportedValue,
+    double tolerance) const
+{
+    if (!d->cam || !d->cam->isOpened())
+        return;
+    if (!differsFromRequested(reportedValue, requestedValue, tolerance))
+        return;
+
+    const auto name = cameraPropertyName(propertyId);
+    LOG_WARNING(
+        d->log,
+        "Requested camera property '{}' value {}, but the backend reports {}. "
+        "Keeping the requested value for future connection attempts.",
+        name,
+        requestedValue,
+        reportedValue);
+}
+
 void Camera::setCamId(int id)
 {
     d->camId = id;
@@ -129,22 +224,47 @@ void Camera::setStartTime(const symaster_timepoint &time)
 void Camera::setResolution(const cv::Size &size)
 {
     d->frameSize = size;
-    d->cam->set(cv::CAP_PROP_FRAME_WIDTH, d->frameSize.width);
-    d->cam->set(cv::CAP_PROP_FRAME_HEIGHT, d->frameSize.height);
+
+    const double requestedWidth = d->frameSize.width;
+    const double requestedHeight = d->frameSize.height;
+    const auto reportedWidth = setCameraProperty(cv::CAP_PROP_FRAME_WIDTH, requestedWidth, false);
+    const auto reportedHeight = setCameraProperty(cv::CAP_PROP_FRAME_HEIGHT, requestedHeight, false);
+    const auto backendReportsDifferentResolution =
+        (reportedWidth && differsFromRequested(*reportedWidth, requestedWidth)) ||
+        (reportedHeight && differsFromRequested(*reportedHeight, requestedHeight));
+
+    if (d->cam && d->cam->isOpened() && backendReportsDifferentResolution) {
+        const auto reportedWidthValue = reportedWidth ? *reportedWidth : requestedWidth;
+        const auto reportedHeightValue = reportedHeight ? *reportedHeight : requestedHeight;
+        LOG_WARNING(
+            d->log,
+            "Requested camera output resolution {}x{}, but the backend reports capture resolution {}x{}. "
+            "Captured frames will be scaled to the requested output resolution.",
+            requestedWidth,
+            requestedHeight,
+            reportedWidthValue,
+            reportedHeightValue);
+    }
 }
 
-int Camera::framerate() const
+double Camera::framerate() const
 {
-    const int capFps = d->cam->get(cv::CAP_PROP_FPS);
-    if (capFps <= 0)
-        return d->fps;
-    return capFps;
+    return isPositiveFinite(d->activeFps) ? d->activeFps : d->fps;
 }
 
-void Camera::setFramerate(int fps)
+void Camera::setFramerate(double fps)
 {
     d->fps = fps;
-    d->cam->set(cv::CAP_PROP_FPS, d->fps);
+    d->activeFps = d->fps;
+
+    const auto reportedFps = setCameraProperty(cv::CAP_PROP_FPS, d->fps, true, 0.5);
+    if (reportedFps && isPositiveFinite(*reportedFps)) {
+        d->activeFps = *reportedFps;
+    } else if (d->cam && d->cam->isOpened()) {
+        const auto currentFps = d->cam->get(cv::CAP_PROP_FPS);
+        if (isPositiveFinite(currentFps))
+            d->activeFps = currentFps;
+    }
 }
 
 cv::Size Camera::resolution() const
@@ -165,7 +285,7 @@ void Camera::setExposure(double value)
         value = 2047;
 
     d->exposure = value;
-    d->cam->set(cv::CAP_PROP_EXPOSURE, value);
+    setCameraProperty(cv::CAP_PROP_EXPOSURE, value);
 }
 
 double Camera::brightness() const
@@ -181,7 +301,7 @@ void Camera::setBrightness(double value)
         value = -100;
 
     d->brightness = value;
-    d->cam->set(cv::CAP_PROP_BRIGHTNESS, value);
+    setCameraProperty(cv::CAP_PROP_BRIGHTNESS, value);
 }
 
 double Camera::contrast() const
@@ -197,7 +317,7 @@ void Camera::setContrast(double value)
         value = 255;
 
     d->contrast = value;
-    d->cam->set(cv::CAP_PROP_CONTRAST, value);
+    setCameraProperty(cv::CAP_PROP_CONTRAST, value);
 }
 
 double Camera::saturation() const
@@ -211,7 +331,7 @@ void Camera::setSaturation(double value)
         value = 255;
 
     d->saturation = value;
-    d->cam->set(cv::CAP_PROP_SATURATION, value);
+    setCameraProperty(cv::CAP_PROP_SATURATION, value);
 }
 
 double Camera::hue() const
@@ -227,7 +347,7 @@ void Camera::setHue(double value)
         value = -100;
 
     d->hue = value;
-    d->cam->set(cv::CAP_PROP_HUE, value);
+    setCameraProperty(cv::CAP_PROP_HUE, value);
 }
 
 double Camera::gain() const
@@ -241,7 +361,7 @@ void Camera::setGain(double value)
         value = 255;
 
     d->gain = value;
-    d->cam->set(cv::CAP_PROP_GAIN, value);
+    setCameraProperty(cv::CAP_PROP_GAIN, value);
 }
 
 int Camera::autoExposureRaw() const
@@ -252,6 +372,7 @@ int Camera::autoExposureRaw() const
 void Camera::setAutoExposureRaw(int value)
 {
     d->autoExposureRaw = value;
+    setCameraProperty(cv::CAP_PROP_AUTO_EXPOSURE, d->autoExposureRaw, true, 0.5);
 }
 
 bool Camera::connect()
@@ -281,23 +402,30 @@ bool Camera::connect()
         LOG_INFO(d->log, "Unable to use preferred camera backend for {} falling back to autodetection.", d->camId);
         ret = d->cam->open(d->camId);
     }
+    if (!ret) {
+        const auto msg = QStringLiteral("Unable to open camera %1.").arg(d->camId);
+        LOG_ERROR(d->log, "{}", msg);
+        fail(msg);
+        return false;
+    }
 
-    d->cam->set(cv::CAP_PROP_FPS, d->fps);
-    d->cam->set(cv::CAP_PROP_AUTO_EXPOSURE, d->autoExposureRaw);
+    d->failed = false;
+    d->lastError.clear();
 
     // set initial defaults
     setPixelFormat(d->captureFormat);
+    setFramerate(d->fps);
+    setAutoExposureRaw(d->autoExposureRaw);
     setExposure(d->exposure);
     setBrightness(d->brightness);
     setContrast(d->contrast);
     setSaturation(d->saturation);
     setHue(d->hue);
+    setGain(d->gain);
 
-    d->cam->set(cv::CAP_PROP_FRAME_WIDTH, d->frameSize.width);
-    d->cam->set(cv::CAP_PROP_FRAME_HEIGHT, d->frameSize.height);
+    setResolution(d->frameSize);
 
     // we are connected now
-    d->failed = false;
     d->connected = true;
 
     // temporary dummy timepoint, until the actual reference starting
@@ -314,6 +442,7 @@ void Camera::disconnect()
     if (d->connected)
         LOG_INFO(d->log, "Disconnected camera {}", d->camId);
     d->connected = false;
+    d->activeFps = d->fps;
 }
 
 QList<CameraPixelFormat> Camera::readPixelFormats()
@@ -348,8 +477,21 @@ void Camera::setPixelFormat(const CameraPixelFormat &pixFmt)
         return;
 
     LOG_INFO(d->log, "Setting pixel format to: {}", pixFmt.fourcc);
-    d->cam->set(cv::CAP_PROP_FOURCC, pixFmt.fourcc);
     d->captureFormat = pixFmt;
+
+    const auto actualFourcc = setCameraProperty(cv::CAP_PROP_FOURCC, pixFmt.fourcc, false);
+    const auto backendReportsDifferentFormat =
+        actualFourcc && (!isPositiveFinite(*actualFourcc) || std::round(*actualFourcc) != pixFmt.fourcc);
+
+    if (d->cam && d->cam->isOpened() && backendReportsDifferentFormat) {
+        LOG_WARNING(
+            d->log,
+            "Requested camera pixel format '{}' ({}), but the backend reports {}. "
+            "Keeping the selected pixel format for future connection attempts.",
+            pixFmt.name,
+            pixFmt.fourcc,
+            *actualFourcc);
+    }
 }
 
 bool Camera::recordFrame(Frame &frame, SecondaryClockSynchronizer *clockSync)
--- a/modules/camera-generic/genericcamerasettingsdialog.ui
+++ b/modules/camera-generic/genericcamerasettingsdialog.ui
@@ -169,18 +169,24 @@
          </widget>
         </item>
         <item row="3" column="1">
-         <widget class="QSpinBox" name="fpsSpinBox">
+         <widget class="QDoubleSpinBox" name="fpsSpinBox">
           <property name="toolTip">
            <string>Frames per second</string>
           </property>
+          <property name="decimals">
+           <number>3</number>
+          </property>
           <property name="minimum">
-           <number>4</number>
+           <double>4.000000000000000</double>
           </property>
           <property name="maximum">
-           <number>30</number>
+           <double>480.000000000000000</double>
+          </property>
+          <property name="singleStep">
+           <double>1.000000000000000</double>
           </property>
           <property name="value">
-           <number>20</number>
+           <double>20.000000000000000</double>
           </property>
          </widget>
         </item>
--- a/modules/camera-generic/camera.h
+++ b/modules/camera-generic/camera.h
@@ -21,6 +21,7 @@
 #define GENERIC_CAMERA_H
 
 #include <QSize>
+#include <expected>
 #include <opencv2/core.hpp>
 
 #include "logging.h"
@@ -64,8 +65,8 @@ public:
     cv::Size resolution() const;
     void setResolution(const cv::Size &size);
 
-    int framerate() const;
-    void setFramerate(int fps);
+    double framerate() const;
+    void setFramerate(double fps);
 
     double exposure() const;
     void setExposure(double value);
@@ -104,6 +105,13 @@ private:
     std::unique_ptr<CameraData> d;
 
     void fail(const QString &msg);
+    std::expected<double, QString> setCameraProperty(
+        int propertyId,
+        double value,
+        bool check = true,
+        // Camera backends might quantize hardware properties before reporting them back.
+        double tolerance = 0.5);
+    void warnIfCameraReportsDifferent(int propertyId, double requestedValue, double reportedValue, double tolerance) const;
 };
 
 #endif // GENERIC_CAMERA_H
--- a/modules/camera-generic/genericcameramodule.cpp
+++ b/modules/camera-generic/genericcameramodule.cpp
@@ -24,6 +24,8 @@
 #include "camera.h"
 #include "genericcamerasettingsdialog.h"
 
+#include <cmath>
+
 SYNTALOS_MODULE(GenericCameraModule)
 
 class GenericCameraModule : public AbstractModule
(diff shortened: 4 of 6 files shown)


modules/camera-generic/camera.cpp
```diff
@@ -111,6 +125,53 @@ void Camera::fail(const QString &msg)
     d->lastError = msg;
 }
 
+double Camera::setCameraProperty(const QString &name, int propertyId, double value, bool *setSucceeded)
```
ximion: Please use `std::expected<double, QString>` or `std::expected<double, void>` here, instead of the `setSucceeded` parameter.

modules/camera-generic/camera.cpp
```diff
@@ -28,6 +29,19 @@
 #include <sys/ioctl.h>
 #include "datactl/frametype.h"
 
+namespace
```
ximion: Please don't open up an anonymous namespace, but declare the helper methods in it later and as `static`.

modules/camera-generic/camera.cpp
ximion: Instead of passing a `name` string around, it would be better to create a mapping table for tha values we actually use, with a sensible fallback in case one is missing (if OpenCV doesn't provide that already). That way, we don't need to remember to pass the name around correctly for logging, which is a thing where you can easily make mistakes.

modules/camera-generic/camera.cpp
```diff
@@ -111,6 +125,53 @@ void Camera::fail(const QString &msg)
...
+        return cameraValue;
+    }
+
+    if (setSucceeded != nullptr)
+        *setSucceeded = true;
+
+    return d->cam->get(propertyId);
+}
+
+void Camera::warnIfCameraReportsDifferent(
```
ximion, line 188: Shouldn't this method be automatically called by `setCameraProperty` if a `check` bool is `true` as parameter? That way, it would be default-on, and the caller doesn't have to remember it. Yet, if we don't want it (e.g. for performance, or if the camera property is write-only) we can set `check = false`.

ximion (changes requested):
Is there a reason for the excessive checking of the selected dimensions? Because OpenCV kind of guarantees that this cannot fail, it will software-scale if needed and you can select literally any dimensions. So, did this fail somehow?

caenrigen:
> Is there a reason for the excessive checking of the selected dimensions? Because OpenCV kind of guarantees that this cannot fail, it will software-scale if needed and you can select literally any dimensions. So, did this fail somehow?

In part this was due to some code iterations. The warning is meant to tell the user when the camera/backend did not actually switch to the requested capture mode, so the requested output will be produced by software scaling from a different native capture resolution. That can matter for image quality, bandwidth, framerate, and debugging camera setup. I want to know if am unnecessarily producing a squished/stretched image in one of the dimensions.

It did not fail, just silently "messing up" the native resolution of the camera.

modules/camera-generic/genericcamerasettingsdialog.ui
```diff
@@ -169,18 +169,24 @@
...
           <property name="minimum">
-           <number>4</number>
+           <double>4.000000000000000</double>
           </property>
           <property name="maximum">
-           <number>30</number>
+           <double>30.000000000000000</double>
+          </property>
+          <property name="singleStep">
+           <double>0.500000000000000</double>
```
ximion: If we are allowing fractional framerates, we should at least not make it to easy to set one and set this to `1.0`, so users don't accidentally go fractional.

caenrigen: agreed

caenrigen:
FYI when initially implementing this PR with Codex help I instructed it to avoid raising errors when setting the camera controls because some cameras do not support at all certain controls (I have one that does not support `hue` and maybe there are some that only support fixed FPS, etc.). This decision makes sense to not prevent people from using their camera with this module.

However, this also triggered a series of CodeRabbit complaints because we still need to decide how exactly to handle requested/supported control value mismatches... This is specially sensitive for FPS given its importance for timings.

I used CodeRabbit suggestions + Codex to fix its complaints, but I do not feel qualified to judge if the correct approaches have been taken.

ximion:
The patch is logically sound, but there are a few stylistic things that I would want to have addressed. I will merge this though, given that it was AI-created, and fix the things up in a follow-up patch, so this PR doesn't sit in the pending queue for even longer :-)

caenrigen:
Thank you!

caenrigen:
NB as of 33385cfe9e08cef4a146420f36fdef21b6f913eb Codex reports

> camera.cpp (line 218) still treats a backend resolution mismatch as non-fatal and even logs that frames “will be scaled”, but recordFrame (line 531) now rejects mismatched frame dimensions instead of resizing. prepare() then starts the stream with requested-size metadata at genericcameramodule.cpp (line 133), and capture fails later after repeated frame errors. Since the final state intentionally does not resize, this should either fail during prepare()/connect() with the actual backend dimensions, or store/use the backend effective resolution for metadata and dimension checks. The current behavior gives users a stale warning and a delayed generic acquisition failure.

ximion:
I don't think Codex understands that the OpenCV API is weird in that you can throw in any dimension and it will always give you the scaled result, without you having to do anything (which is actually *terrible* - I would much rather ask the camera what sizes it actually supports and then use those, which is another case for a libcamera-based Syntalos module...)

caenrigen:
> I don't think Codex understands that the OpenCV API is weird

upsss 🙈 

> (which is actually terrible - I would much rather ask the camera what sizes it actually supports and then use those, which is another case for a libcamera-based Syntalos module...)

Actually I have been working most of the day (more like supervised vibe-coding...) on a C++ `camera-v4l2` module for Syntalos using only v4l2 (ffmpeg for decoding). This time I started with a longer design-ish document and have been pushing Codex to review it, find bugs, compare with features in camera-tis, adapt the GUI to my liking, add some quirks to make it work for the particular camera I intend to use (Logitech Webcam C930e).

Any thoughts on a PR into Syntalos with the result of these efforts?

ximion:
Oh nice! I have a libcamera-based prototype as well (just a "how hard can it be to use modern APIs?" thing), let me commit that so you can have a look, maybe it's useful to expand upon!

Generally, as long as a module isn't completely unmaintainable code, I'd add it to Syntalos. The only caveat is that modules from 3rd-parties will also need to be maintained by them to some extent - I promise to keep stuff building and to adjust modules to any API changes, but since for many I won't even have the hardware, that's kind of the extent of it.

All that said, having your eye tracking Python module in Syntalos would also be totally neat, so it would be easily accessible for others to use :-)

ximion:
Actually, having looked into it, the libcamera module might give you what you need even easier: I just plumbed the powerline frequence setting through libcamera to the V4L2 API, because looking at the constant flickering made my head hurt. It's a lot more deterministic and easy than with OpenCV, and may be an option for you.

Libcamera also kinda offers access to the raw device node exactly for this purpose. Of course, all of these things will make life hell for anyone who wants to attempt a Windows port of Syntalos :-P But it's quite smooth API on Linux (and is orders of magnitude more precise than the OpenCV API, I just measured time jitter on the device, and, damn...).

caenrigen:
> Oh nice! I have a libcamera-based prototype as well (just a "how hard can it be to use modern APIs?" thing), let me commit that so you can have a look, maybe it's useful to expand upon!

Interesting! I will give it a try today 😀

> All that said, having your eye tracking Python module in Syntalos would also be totally neat, so it would be easily accessible for others to use :-)

😇 As a start might be easier to have a docs page with links to "3rd party modules" that points to my repo (NB I still need to move my modules under a Github Org managed by the lab). And the time sync is not yet properly finalised.

> Actually, having looked into it, the libcamera module might give you what you need even easier: I just plumbed the powerline frequence setting through libcamera to the V4L2 API, because looking at the constant flickering made my head hurt.

Progress out of necessity haha

> It's a lot more deterministic and easy than with OpenCV, and may be an option for you.

On one hand, this is promising
 
> Libcamera also kinda offers access to the raw device node exactly for this purpose.

On the other hand, what is the advantage of using an extra library on top of v4l2 when several controls still need to be wired via lower level "hooks"? As most of the time it is a pros and cons game

> Of course, all of these things will make life hell for anyone who wants to attempt a Windows port of Syntalos :-P But it's quite smooth API on Linux (and is orders of magnitude more precise than the OpenCV API, I just measured time jitter on the device, and, damn...).

I am not planning to switch to Window anytime soon, and hopefully never 😂😅

---

I will iterate on my v4l2 module and share it via a PR, we can perhaps "merge" the learnings from both v4l2 and libcamera

ximion:
> 😇 As a start might be easier to have a docs page with links to "3rd party modules" that points to my repo (NB I still need to move my modules under a Github Org managed by the lab). And the time sync is not yet properly finalised.

I have some rough ideas to allow downloadable modules, but that will be a bit painful due to binary incompatibilities and different Syntalos versions for mlink modules, and due to different Python versions for Python modules (depending on what people use). And I don't want a frustrating experience where people fetch a module and it immediately doesn't work at all due to dependency issues. Still something to be worked on (there are solutions, but all of them are quite complex, and the Syntalos module ecosystem is small - shipping them with the project directly is easiest right now, alongside having people who know their stuff ship them.

> On the other hand, what is the advantage of using an extra library on top of v4l2 when several controls still need to be wired via lower level "hooks"? As most of the time it is a pros and cons game

libcamera still provides a lot of abstractions that I don't have to write, which is exceptionally nice. It also works (in theory) with non-V4L2 stuff. But most importantly, it provides the image processing support for certain modern embedded cameras, e.g. some stuff used on Raspberry Pis & Co. I expect the use of those to also increase. And at that point, a lot of extra processing will be needed on top of V4L2, which libcamera provides. So, using libcamera is future-proofing the module and allowing it to access more cameras (as well as having a friendlier API).

Going through to V4L2 for some stuff that isn't abstracted away (yet) actually feels fine - and we don't need to use crazy heuristics like with the OpenCV module for that, libcamera is nicer there :-)

> I will iterate on my v4l2 module and share it via a PR, we can perhaps "merge" the learnings from both v4l2 and libcamera

I am not yet 100% sure what you need, but chances are that you can just replace the rather basic settings GUI I made for libcamera with yours, and plumb through the remaining V4L interfaces that libcamera doesn't provide yet. That result should be quite powerful. Not sure though if you may need to fetch a newer libcamera from Debian 13 backports though.

> add some quirks to make it work for the particular camera I intend to use (Logitech Webcam C930e)

I have a C925e, quite a good device! (although also not the cheapest when I bought it during the pandemic...)
Generally I am against weird quirk fixes, as they do tend to cause issues, but, for these devices there sometimes really is no other way. This is why even the OpenCV camera module had a dedicated "Quirks" panel with instructions on which values to set when...

If we can have Syntalos do the right thing and not offload a "find the magic values!" task to the user, we should absolutely do that - but if we can't, then the UI should offer some explanations so the "find magic values" quest will at least be as short as possible for the user.

When working with webcam-class devices, you learn how nice of a thing GenICam really is, and afterwards definitely appreciate it more :-)

caenrigen:
> Interesting! I will give it a try today 😀

From a quick test, `camera-lc` seems to work for me too

> shipping them with the project directly is easiest right now, alongside having people who know their stuff ship them.

Sounds reasonable. I can ping you when I feel my modules are complete and stable enough.

> libcamera still provides a lot of abstractions that I don't have to write, which is exceptionally nice. It also works (in theory) with non-V4L2 stuff. But most importantly, it provides the image processing support for certain modern embedded cameras, e.g. some stuff used on Raspberry Pis & Co. I expect the use of those to also increase. And at that point, a lot of extra processing will be needed on top of V4L2, which libcamera provides. So, using libcamera is future-proofing the module and allowing it to access more cameras (as well as having a friendlier API).

Understood

> but chances are that you can just replace the rather basic settings GUI I made for libcamera with yours, and plumb through the remaining V4L interfaces that libcamera doesn't provide yet. That result should be quite powerful.

I will give this a try. The main control I was missing is the auto-focus and manual focus controls

> Not sure though if you may need to fetch a newer libcamera from Debian 13 backports though.

It seemed to work with libcamera-dev 0.4 but just in case I installed libcamera-dev 0.7 from backports

> I have a C925e, quite a good device!

Does it also support auto and manual focus? Maybe share your:

```bash
sudo apt install v4l-utils
v4l2-ctl -d /dev/video0 --list-ctrls-menus
v4l2-ctl -d /dev/video0 --list-formats-ext
```

> This is why even the OpenCV camera module had a dedicated "Quirks" panel with instructions on which values to set when...

For that particular one (auto exposure mode), I don't think it had to be called a "quirk", auto/manual exposure seems like a reasonable feature for webcams to have. Idem for auto/manual focus.

Of course it becomes a quirk of the particular camera models (or their drivers) if the values for that v4l2 parameter does not correspond to a very similar physical effect.

> If we can have Syntalos do the right thing and not offload a "find the magic values!" task to the user, we should absolutely do that - but if we can't, then the UI should offer some explanations so the "find magic values" quest will at least be as short as possible for the user.

Agreed

> When working with webcam-class devices, you learn how nice of a thing GenICam really is, and afterwards definitely appreciate it more :-)

I am getting there 😂😂😂

At the same time, so many experiments are perfectly fine with a cheap-ish webcam-like device (if the software supports it well 😅)

---

> Generally I am against weird quirk fixes, as they do tend to cause issues, but, for these devices there sometimes really is no other way.

Agreed, tends to not be elegant but makes a device usable 🤷🏼 

For my C930e I run into 3 quirks (there is some chance that my camera(s) have defects, old devices I have in the lab...):

Quirk 1: Turning off an auto-control did not automatically (re)set the physical parameter to the value that the corresponding manual control reports.

E.g., You change exposure from auto to manual, now the manual exposure control is active again but its value does not match what the camera output is displaying. Reading the manual exposure and setting it again fixed it. In your `camera-lc`, `libcamera` seems to do this step under the hood (or your code does, i did not look into it yet).

The same procedure is needed for transitioning from auto-focus to manual focus, but that is where I run into:

Quirk 2: Setting the manual focus right after disabling the auto-focus does work. Even though the driver reports the manual focus control being writable and accepting the parameter the physical focus does not change. The only way is to wait ~50 ms and then set the manual focus (to any value) to bring the driver/UI in sync with the physical focus. In my v4l2 prototype, I added a delay spin box quirk to each auto-control to deal with this.

Quirk 3: C930e seems to persist some settings when disconnecting the camera (just in case I reboot the machine to decrease the chance that this is a driver-based persistence, might still be if it caches control values on disk...). After reboot/replug, the driver reports the same auto-focus (off) and the same manual-focus (previous set value).

The annoying part is that both of these settings are out of sync and setting new values to the manual control does nothing to the image (physically the camera seems to be in auto-mode and ignoring the manual values). Just applying again the focus to manual does not seem to fix it. It requires to toggle the auto-focus on and off again for it to start working properly. So far i did not implement anything to deal with this, but it seems required, otherwise any machine reboot will silently break the image since there is no programmatic interface to check this (I did not check if there is any lower level logging that could spot it)...

Since I would need to plumb the auto focus into `camera-lc`, I also need to handle these quirks... Acceptable? Thoughts? Wanna check the behavior of your C925e?

caenrigen:
From looking at the controls provided by my C930e but missing in `camera-lc`:

Gamma was just missing in your implementation, PR submitted: #161 

Relevant ones that exist in `libcamera` but UVC pipeline does not implement them. Most likely because UVC/webcam devices often do not generally fit its semantics cleanly:

- Hue
    - Likely because V4L2/UVC ranges/units are not cleanly normalized to libcamera’s degrees model
- Sharpness
    - UVC sharpness is usually a device processing-unit value with arbitrary scaling, so normalizing it is fuzzy.
- Auto white balance temperature / White balance temperature
    - UVC commonly exposes this directly as Kelvin plus auto-WB. libcamera models WB more like an ISP pipeline: AwbEnable, ColourGains (red and blue gains, green as reference), and reported ColourTemperature. ColourTemperature is generated output-only, and UVC does not map WB controls.
- Auto focus / manual focus (relative/absolute)
    - libcamera LensPosition is in dioptres, while UVC focus absolute is generally raw device units, and focus relative is a step command. Without calibration, libcamera cannot honestly expose raw UVC focus as dioptres.

Not a big priority (though wiring them will likely be cheap once the other V4L2 controls are implemented in a generic manner):

- Backlight compensation (generally only applicable to auto-exposure)
- Pan
- Tilt
- Zoom

ximion:
For arbitrary values where we don't know limits, we could cheat and just give the user a spinbox to insert any value they want. Not great UX-wise, but also the most honest representation of the control that we can provide then.

```
$ v4l2-ctl -d /dev/video0 --list-ctrls-menus

User Controls

                     brightness 0x00980900 (int)    : min=0 max=255 step=1 default=128 value=128 flags=has-min-max
                       contrast 0x00980901 (int)    : min=0 max=255 step=1 default=128 value=128 flags=has-min-max
                     saturation 0x00980902 (int)    : min=0 max=255 step=1 default=128 value=128 flags=has-min-max
        white_balance_automatic 0x0098090c (bool)   : default=1 value=1
                           gain 0x00980913 (int)    : min=0 max=255 step=1 default=0 value=0 flags=has-min-max
           power_line_frequency 0x00980918 (menu)   : min=0 max=2 default=2 value=2 (60 Hz)
                                0: Disabled
                                1: 50 Hz
                                2: 60 Hz
      white_balance_temperature 0x0098091a (int)    : min=2000 max=6500 step=1 default=4000 value=4000 flags=inactive, has-min-max
                      sharpness 0x0098091b (int)    : min=0 max=255 step=1 default=128 value=128 flags=has-min-max
         backlight_compensation 0x0098091c (int)    : min=0 max=1 step=1 default=0 value=0 flags=has-min-max

Camera Controls

                  auto_exposure 0x009a0901 (menu)   : min=0 max=3 default=3 value=3 (Aperture Priority Mode)
                                1: Manual Mode
                                3: Aperture Priority Mode
         exposure_time_absolute 0x009a0902 (int)    : min=3 max=2047 step=1 default=250 value=250 flags=inactive, has-min-max
     exposure_dynamic_framerate 0x009a0903 (bool)   : default=0 value=1
                   pan_absolute 0x009a0908 (int)    : min=-36000 max=36000 step=3600 default=0 value=0 flags=has-min-max
                  tilt_absolute 0x009a0909 (int)    : min=-36000 max=36000 step=3600 default=0 value=0 flags=has-min-max
                 focus_absolute 0x009a090a (int)    : min=0 max=250 step=5 default=0 value=0 flags=inactive, has-min-max
     focus_automatic_continuous 0x009a090c (bool)   : default=1 value=1
                  zoom_absolute 0x009a090d (int)    : min=100 max=500 step=1 default=100 value=100 flags=has-min-max


$ v4l2-ctl -d /dev/video0 --list-formats-ext
ioctl: VIDIOC_ENUM_FMT
        Type: Video Capture

        [0]: 'YUYV' (YUYV 4:2:2)
                Size: Discrete 640x480
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 160x90
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 160x120
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 176x144
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 320x180
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 320x240
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 352x288
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 432x240
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 640x360
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 800x448
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 800x600
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 864x480
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 960x720
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 1024x576
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 1280x720
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 1600x896
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 1920x1080
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 2304x1296
                        Interval: Discrete 0.500s (2.000 fps)
                Size: Discrete 2304x1536
                        Interval: Discrete 0.500s (2.000 fps)
        [1]: 'MJPG' (Motion-JPEG, compressed)
                Size: Discrete 640x480
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 160x90
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 160x120
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 176x144
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 320x180
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 320x240
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 352x288
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 432x240
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 640x360
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 800x448
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 800x600
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 864x480
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 960x720
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 1024x576
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 1280x720
                        Interval: Discrete 0.017s (60.000 fps)
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 1600x896
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
                Size: Discrete 1920x1080
                        Interval: Discrete 0.033s (30.000 fps)
                        Interval: Discrete 0.042s (24.000 fps)
                        Interval: Discrete 0.050s (20.000 fps)
                        Interval: Discrete 0.067s (15.000 fps)
                        Interval: Discrete 0.100s (10.000 fps)
                        Interval: Discrete 0.133s (7.500 fps)
                        Interval: Discrete 0.200s (5.000 fps)
```

> At the same time, so many experiments are perfectly fine with a cheap-ish webcam-like device (if the software supports it well 😅)

The Miniscope project once made an ultra-cheap webcam for experiments, but that's discontinued now. Hopefully we will get more standardization in this field for consumer hardware, libcamera and the fact that Android exists will probably actually help a lot for this.


> Quirk 3: C930e seems to persist some settings when disconnecting the camera (just in case I reboot the machine to decrease the chance that this is a driver-based persistence, might still be if it caches control values on disk...). After reboot/replug, the driver reports the same auto-focus (off) and the same manual-focus (previous set value).

Quite common for those devices, and I kinda like it :-) - but it's also the reson why Syntalos sets whatever the user saved again every time, instead of assuming that the value it read back is sufficient.

There is a chance that what you implement for your camera won't work for all of them, and that we break your usecase as soon as the next guy comes along and tries to fix it for their camera - that is, unless there is enough standardization for a control. Usually you can test with `qv4l2` and see what it display and how it reads/implements the values. Doing what that tool does was so far relatively reliable and portable across devices.

caenrigen:
> For arbitrary values where we don't know limits, we could cheat and just give the user a spinbox to insert any value they want. Not great UX-wise, but also the most honest representation of the control that we can provide then.

I have not come across such controls yet. Maybe keep it simple and wait for someone to need them.

> ```
> $ v4l2-ctl -d /dev/video0 --list-ctrls-menus
> ```

Your C925e has the same controls as my C930e

> The Miniscope project once made an ultra-cheap webcam for experiments, but that's discontinued now.

🥲 

> There is a chance that what you implement for your camera won't work for all of them, and that we break your usecase as soon as the next guy comes along and tries to fix it for their camera - that is, unless there is enough standardization for a control.

Fair. As a start we can have clear comments with device model for each quirk.

> Usually you can test with qv4l2 and see what it display and how it reads/implements the values. Doing what that tool does was so far relatively reliable and portable across devices.

Cool, I was not aware of this tool, thanks!

ximion:
> Maybe keep it simple and wait for someone to need them.

I would definitely strongly suggest to only implement what you currently need and can test ;-)

> Your C925e has the same controls as my C930e

Yeah, I was surprised, I had no idea it could do manual focus - but I also haven't tested that yet.

You can still buy the MiniCam: https://www.labmaker.org/collections/neuroscience/products/minicam - but the project is listed as discontinued, so, getting a GenICam camera may be better overall (especially since there are some vendors who even offer cheaper devices...)

> Cool, I was not aware of this tool, thanks!

It's my go-to for testing V4L devices :-)

PR #159, opened 2026-05-22
videotransform: Add Mirror and clockwise Rotate transforms

caenrigen:


ximion:
> MirrorTransform axis mapping is consistent (no inversion bug in current implementation)

What kind of review comment is that??? "Code too good, can't merge" - AI is very strange sometimes...

caenrigen:
> > MirrorTransform axis mapping is consistent (no inversion bug in current implementation)
> 
> What kind of review comment is that??? "Code too good, can't merge" - AI is very strange sometimes...

If we ask it to find things, it will always regurgitate something 😂😂😂

I tried fixing its nitpick but it just makes code more bulky with little gain

modules/videotransform/videotransform.cpp
```diff
@@ -505,6 +506,154 @@ bool ScaleTransform::needsIndependentCopy() const
...
+
+RotateTransform::RotateTransform(int degrees)
+    : VideoTransform(),
+      m_degrees(normalizeRotationDegrees(degrees))
+{
+}
+
+QString RotateTransform::name() const
+{
+    return QStringLiteral("Rotate Frames");
```
ximion:
```suggestion
    return QStringLiteral("Rotate");
```

Nitpick, but matches the naming of the other modules ("Frames" is implied, as that is the only input)

modules/videotransform/videotransform.cpp
```diff
@@ -505,6 +506,154 @@ bool ScaleTransform::needsIndependentCopy() const
...
+MetaSize RotateTransform::resultSize()
+{
+    if (m_degrees == 90 || m_degrees == 270)
+        return {m_originalSize.height, m_originalSize.width};
+    return m_originalSize;
+}
+
+void RotateTransform::process(cv::Mat &image)
+{
+    cv::Mat outMat;
```
ximion: Nitpick: You don't need that, it should be safe to act on that image directly, so `cv::transpose(image, image)` should work. That can potentially save one extra memory copy (but won't necessarily, depends on the algorithm OpenCV uses).

modules/videotransform/videotransform.cpp
```diff
@@ -505,6 +506,154 @@ bool ScaleTransform::needsIndependentCopy() const
...
+
+MirrorTransform::MirrorTransform(Axis axis)
+    : VideoTransform(),
+      m_axis(axis)
+{
+}
+
+QString MirrorTransform::name() const
+{
+    return QStringLiteral("Mirror Frames");
```
ximion:
```suggestion
    return QStringLiteral("Mirror");
```

modules/videotransform/videotransform.cpp
```diff
@@ -505,6 +506,154 @@ bool ScaleTransform::needsIndependentCopy() const
...
+}
+
+QIcon MirrorTransform::icon() const
+{
+    if (m_axis == Axis::X)
+        return QIcon::fromTheme("object-flip-horizontal");
+    return QIcon::fromTheme("object-flip-vertical");
```
ximion, line 606: Haha, nice touch! Just keep in mind that this icon won't update "live" - but I think that's okay here.

modules/videotransform/videotransform.cpp
```diff
@@ -505,6 +506,154 @@ bool ScaleTransform::needsIndependentCopy() const
...
+
+    auto formLayout = new QFormLayout(parent);
+    formLayout->addRow(QStringLiteral("Axis:"), axisSelector);
+    parent->setLayout(formLayout);
+}
+
+void MirrorTransform::process(cv::Mat &image)
+{
+    cv::Mat outMat;
+    cv::flip(image, outMat, m_axis == Axis::X ? 1 : 0);
```
ximion:
```suggestion
    cv::flip(image, image, m_axis == Axis::X ? 1 : 0);
```

Here, the extra outMap should definitely not be needed...

ximion (changes requested):
Basically just nitpicks :-)

modules/videotransform/videotransform.cpp
caenrigen: Codex efforts haha

modules/videotransform/videotransform.cpp
caenrigen: NB I fixed Scale and Crop too

modules/videotransform/videotransform.cpp
```diff
@@ -505,6 +506,150 @@ bool ScaleTransform::needsIndependentCopy() const
...
+
+void RotateTransform::fromVariantHash(const QVariantHash &settings)
+{
+    m_degrees = normalizeRotationDegrees(settings.value("rotation_degrees", 90).toInt());
+}
+
+bool RotateTransform::needsIndependentCopy() const
+{
+    return false;
+}
```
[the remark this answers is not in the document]
ximion: @caenrigen This is a valid review comment - you can make `needsIndependentCopy` return true unconditionally (simply delete it, true is the default), or you can check if you are actually performing any transform, and only then request a copy (if the image is passing through unchanged, you won't need a copy).

The surrounding code will then make sure that the image is only copied *once* at most for all transforms.

ximion (approved):
LGTM

PR #161, opened 2026-06-02
camera-lc: Add gamma control

caenrigen:


ximion:
Straightforward patch, no notes, thank you! :-)

PR #162, opened 2026-06-02
Camera lc queue controls

caenrigen:
NB includes #161 commit

### DIFF
--- a/modules/camera-lc/lccamera.cpp
+++ b/modules/camera-lc/lccamera.cpp
@@ -21,11 +21,14 @@
 
 #include <algorithm>
 #include <array>
+#include <cerrno>
 #include <condition_variable>
+#include <cstdint>
 #include <fcntl.h>
 #include <linux/videodev2.h>
 #include <map>
 #include <mutex>
+#include <optional>
 #include <ostream>
 #include <queue>
 #include <streambuf>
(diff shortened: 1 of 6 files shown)


ximion:
I understand the added features, but can you explain why that complex snapshotting and thread synchronization logic is needed? The libcamera bits should be threadsafe even at runtime already via `libcamera::ControlList` changes, and even the V4L2 bits should be fine if they are used reentrant (and not fully threadsafe). That is the case here, as only one thread ever modifies properties of the camera simultaneously.

So, tl;dr, can you explain why that complexity is needed? Maybe I am just not seeing it...

caenrigen:
Sorry! I forgot to fill in a description 🤦🏼‍♂️

> can you explain why that complexity is needed?

You are right, the complexity is not needed. Just a result of my lack of expertise and rounds of fix-and-review via Codex getting out of hand while I was trying to make sure the module did not have critical flaws. Sometimes it so capable/useful, but trying to design and iterate with it on some issue/feature has not yet given me very good results (likely due to my lack of knowledge and experience with C++, with Python it goes better, I feel more competent at judging and guiding its output).

Codex reported a "data race on the live control fields":

> Live control mutation has no threading boundary. The settings dialog leaves the controls enabled while running, and slots write directly into LcCamera fields (lcsettingsdialog.cpp (line 201)), while the capture thread reads those same fields during request requeue (lccamera.cpp (line 880), lccamera.cpp (line 622)). In C++ this is a data race and it also makes future live-control behavior hard to reason about. Either disable all mutable controls while running, or add an explicit thread-safe control state/command queue like the newer V4L2 module direction.

I guess that is not really any big deal when users tweak parameters.

Closing this, sorry for the slop and wasting your time!

ximion:
> I guess that is not really any big deal when users tweak parameters.

It even is explicitly designed to be allowed ;-)
And as far as I understand the libcamera API, this is perfectly fine.
However, the properties themselves should have been atomic to ensure absolute correctness, which they are now :-)

caenrigen:
Cool! 💪🏼

PR #164, opened 2026-06-08
trivial: Gitignore clangd and mimalloc download caches

caenrigen:
Just a few things that I find myself wanting to ignore in the repo over and over again, I hope you don't mind

the clangs is from a VS Code plugin

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
 build/
 *.user
 __pycache__
+
+# IDE clangd cache
+.cache/clangd
+# mimalloc download
+**/packagecache/


ximion:
Hmm... The `packagecache` ignore is definitely wrong, that one should be auto-ignored... Meson takes care of that, and I would rather like to know why that fails, instead of papering over it.

And can't you simply drop a `.gitignore` file into `.cache` that just contains `*`? That should fix that one as well...

caenrigen:
Ohhh, good tip, I will do that! Thanks.

Maybe the malloc package is from previous attempts in my local repo, will delete it and check again in fresh git clone

ximion:
> Maybe the malloc package is from previous attempts in my local repo, will delete it and check again in fresh git clone

That could be! But Meson should actually already set this one to be ignored, so this is a bit strange. Deleting it to see if it shows up again is actually a safe choice here though :-)

caenrigen:
Indeed just deleting the files was enough, however I was having also some strange modified imgui files.
Cloning the repo fresh solved that too.

Glad that this PR is not needed :) Thank you!

PR #167, opened 2026-06-11
camera-arv: Display status message on FPS mismatch

### LINKED ISSUE #166: camera-arv: missing slow FPS warning
@ximion I realized by accident that the `RGB 8` / `BGR 8` formats have some streaming performance issues, can't reach 60 FPS, while, e.g., `Bayer RG 8` can easily do 120 FPS on my Basler.

It would be pretty handy to have a module info message (similar to other `camera-*` modules) when the requested target FPS is not achieved.

I only noticed it because I have a second camera that cannot do color and its recorded video was going at a different speed compared to the colored camera.

Image

---

Do you have any real world experience of what the max FPS I can expect to achieve for a USB-3 colored camera GigE (screenshot above) and what should be reasonable FPS for the video recorder to keep up (in real time)?

caenrigen:
Fixes #166

### DIFF
--- a/modules/camera-arv/araviscameramodule.cpp
+++ b/modules/camera-arv/araviscameramodule.cpp
@@ -29,6 +29,8 @@
 #undef _SYTMP_QT_SIGNALS_DEFINED
 #endif
 
+#include <atomic>
+
 #include "datactl/frametype.h"
 #include "configwindow.h"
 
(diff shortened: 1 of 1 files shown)


caenrigen:
NB Codex-implemented

ximion:
This looks fine overall, I would have some nitpicks though (the coding style could be better, and there's some odd variable naming choices, and the order of operations is also not ideal (the ready-lock should happen later). Do you want me to comment those, so you can fix them, or should I resolve those issues in a follow-up commit?

caenrigen:
Just push directly, guess it takes you less time to code it than to comment and explain etc.

ximion:
> Just push directly, guess it takes you less time to code it than to comment and explain etc.

It is faster, but how will you eventually learn C++? :wink: 
All changes are fairly minor, the only logic change is to move the wait barrier, which will only gain us a few ns in better time alignment anyway (or none, depending on what code the compiler generates - compilers got really smart with this logic).

ximion:
One issue for sure: The fixed 2 fps tolerance won't work - at higher framerates ~200fps, that's basically normal fluctuation, while at lower framerates like 25 fps it's an actual reason for concern.

So I will make that a percent of the expected rate instead. Also, I will reduce nesting in the code a bit, it's already a bit hard to read with all the lambdas.

caenrigen:
Thank you for taking care of it!

> It is faster, but how will you eventually learn C++? 😉

🙈

I am hoping to start learning Rust instead at some point 😅

caenrigen:
> One issue for sure: The fixed 2 fps tolerance won't work - at higher framerates ~200fps, that's basically normal fluctuation, while at lower framerates like 25 fps it's an actual reason for concern.

you might want to apply the percent/based logic to the other `camera-*` modules too, the 2 fps was inspired from the other modules.

Some modern devices might be able to provide 120/240 fps too

Issue #163, opened 2026-06-03
camera-lc: Failed to set controls


caenrigen:
After 10s or so of streaming I get (and the stream gets broken):

```
15:41:19.853940 I syntalos:engine: Preparing 'Camera'...
15:41:19.864393 I syntalos:libcamera: [27:34:44.879552006] [502277]  INFO Camera camera.cpp:1216 configuring streams: (0) 1920x1080-MJPEG/Rec709/sRGB/Rec601/Limited
15:41:19.869662 I syntalos:engine: Module 'Camera' prepared in 14 msec
...
15:41:20.136492 I syntalos:engine: Running...
15:41:28.052526 E CameraManager:libcamera: [27:34:53.067674690] [502300] ERROR V4L2 v4l2_device.cpp:419
/dev/video0[39:cap]: Unable to set control 0x00980913: Numerical result out of range
15:41:28.052540 E CameraManager:libcamera: [27:34:53.067700096] [502300] ERROR UVC uvcvideo.cpp:441
Failed to set controls: 2
15:41:28.585395 E CameraManager:libcamera: [27:34:53.600544397] [502300] ERROR V4L2 v4l2_device.cpp:419
/dev/video0[39:cap]: Unable to set control 0x00980902: Numerical result out of range
15:41:28.585400 E CameraManager:libcamera: [27:34:53.600565707] [502300] ERROR UVC uvcvideo.cpp:441
Failed to set controls: 3
15:41:30.219357 E CameraManager:libcamera: [27:34:55.234502528] [502300] ERROR V4L2 v4l2_device.cpp:419
/dev/video0[39:cap]: Unable to set control 0x009a0902: Numerical result out of range
15:41:30.219365 E CameraManager:libcamera: [27:34:55.234529771] [502300] ERROR UVC uvcvideo.cpp:441
Failed to set controls: 1
15:41:31.317780 E CameraManager:libcamera: [27:34:56.332930860] [502300] ERROR V4L2 v4l2_device.cpp:419
/dev/video0[39:cap]: Unable to set control 0x009a0901: Numerical result out of range
15:41:31.317788 E CameraManager:libcamera: [27:34:56.332954155] [502300] ERROR UVC uvcvideo.cpp:441
Failed to set controls: 6
```

It seems that `loadSettings` is passing invalid control values to `libcamera`.

Codex says:

[...]

PR #168, opened 2026-06-11
camera-lc: Log getFrame timeout

caenrigen:
A 5s timeout is worth logging. It helps to spot the issue I mentioned https://github.com/syntalos/syntalos/issues/163#issuecomment-4679924934

Additionally you might want to consider handling this a bit differently because if this timeout repeats itself, it makes no sense to way till it happens 32 times before actually raising an error 😅 (i know this module is recent and for now it has mainly generic error handling)


ximion:
I'll add some stuff to this post-merge, to make this more than just a log message.

PR #175, opened 2026-06-25
python: Add set_status_message function

caenrigen:
I have a few use cases in which I would like to have control over the status message.

This is Codex-implemented draft, I hope it is clean and simple enough.

Idk if it makes more sense to be a method on the SyntalosLink object or on the global `syntalos_mlink`. The current implementation follows the `syntalos_mlink.raise_error` pattern.

### DIFF
--- a/src/python/pysy-mlink.cpp
+++ b/src/python/pysy-mlink.cpp
@@ -596,6 +596,11 @@ public:
         m_slink->setState(ModuleState::IDLE);
     }
 
+    void setStatusMessage(const std::optional<std::string> &message)
+    {
+        m_slink->setStatusMessage(message.value_or(std::string()));
+    }
+
     /**
      * Returns true if the experiment is currently running.
      */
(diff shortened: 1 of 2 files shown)


caenrigen:
@ximion I will wait for your thoughts before addressing any of the rabbit comments

caenrigen:
@ximion could you have a look? I hope you are doing well 🙂

No rush, just checking if you might have missed it

src/python/pysy-mlink.cpp
```diff
@@ -1340,6 +1345,17 @@ PYBIND11_MODULE(syntalos_mlink, m)
         "\n"
         ":param message: Human-readable error description.");
 
+    m.def(
+        "set_status_message",
+        set_status_message,
+        py::arg("message") = py::none(),
+        "Set the GUI-visible module status message.\n"
+        "\n"
+        "May contain Qt rich text.\n"
```
ximion:
```suggestion
        "May contain limited HTML rich text if it starts with '<html>' and must be as short as possible. Long messages will be truncated.\n"
```

ximion (changes requested):

src/python/pysy-mlink.cpp
```diff
@@ -1340,6 +1345,17 @@ PYBIND11_MODULE(syntalos_mlink, m)
         "\n"
         ":param message: Human-readable error description.");
 
+    m.def(
+        "set_status_message",
```
ximion: This does not belong here. Please move it into the `SyntalosLink` class (`py::class_<PySyLinkManager>(m, "SyntalosLink")`). This could even be a property, but since it's a fire-and-forget status message, having just the setter function also has something going for it.

ximion (changes requested):

ximion:
I was on vacation ;-)

It looks fine in general, funny that I forgot to fully hook this up, apparently... In any case, this should not be a free-standing function, we tried to get away from those as much as possible. Since this is a feature for full-on Python modules only as well, putting it into the `SyntalosLink` singleton interface class is the thing to do :-)

caenrigen:
> I was on vacation ;-)

I assumed so indeed, hope you had a great vacation! 😄

---

I think it should be fixed now

ximion:
Nice, thank you! Yes, the vacation was very nice and very needed :-)

Issue #176, opened 2026-07-03
Board files are saved/loaded with non deterministic structure


caenrigen:
@ximion A follow up from https://github.com/syntalos/syntalos/discussions/28#discussioncomment-15320636

i put my boards in a git repository and added a pre-commit hook thay automatically unpacks the board into a separate git-tracked folder to easily keep track of the modules settings changes.

I found out that simply loading and saving again the same board (without any changes) completely changes the internal directory structure which sadly makes it very hard to meaningfully keep track of the changes in my boards.

From a quick inspection via Codex, it seems that the modules directories inside the archive are named 0, 1, 2, 3, ..., 10, etc.. However when loading the board back Syntalos uses a simple sort() that sorts strings instead (the names of the dirs) which ends up sorting like 0, 1, 10, 11, 12, 2, etc.

I think sorting based on the numeric value instead will solves it. Is such PR welcome? It should be fully backwards compatible.

More broadly, what are your thoughts on making the boards load/save as deterministic as possible? There are potentially some more related issues. 

Toml serialisation seemed to rely on hashing for keys order but certain hashing functions are process-launch-dependent (I.e. Different output on every program launch).

Another case is that removing a module from the board will change the numbering of the directories. Preserving module "export index" would be handy

ximion:
> From a quick inspection via Codex, it seems that the modules directories inside the archive are named 0, 1, 2, 3, ..., 10, etc.. However when loading the board back Syntalos uses a simple sort() that sorts strings instead (the names of the dirs) which ends up sorting like 0, 1, 10, 11, 12, 2, etc.

Yeah, this is a bug.
The file format is not meant to be git-diffable and there are no determinism guarantees, *however* besides the data change, saving a file with the same Syntalos version should produce the exact same output still. This isn't for your usecase, but simply because it is better practice, reduces bugs, and we do want reproducible data files.

Removing a module changing ordering is fine though, as soon as you make changes, there aren't any guarantees. The TOML output should be identical though, that one is strange!

caenrigen:
> Removing a module changing ordering is fine though, as soon as you make changes, there aren't any guarantees.

git is able to spot some of the file renaming, fine for now, I mainly wanted to track small module config changes.

> The file format is not meant to be git-diffable

I would argue that it should be versioning-friendly to the extent possible. For example it helped me spot the `lossless` bug.

You already had the need of a dedicated script for flattening the project file for easy comparison, though I suppose that is intended mainly for development workflows.

[...]

PR #177, opened 2026-07-04
projectfile: Sort syct entries numerically when loading

caenrigen:
Follow up from #176, partial fix, Codex-implemented


ximion:
I implemented a slightly different solution that addresses this from the reader and from the writer side.

caenrigen:
> I implemented a slightly different solution that addresses this from the reader and from the writer side.

Looks neat, thank you!

PR #183, opened 2026-07-10
trivial: Fix pyenv success message

caenrigen:
Pressing other keys does not really work, as my colleague found out 😂

### DIFF
--- a/src/fabric/pyvenvmanager.cpp
+++ b/src/fabric/pyvenvmanager.cpp
@@ -297,7 +297,7 @@ auto createPythonVirtualEnv(const QString &venvName, const QString &requirements
         << "    if [ $? -ne 0 ]\n"
         << "    then\n"
         << "        echo \"\"\n"
-        << "        read -p \"Command failed to run. Press enter to exit.\"\n"
+        << "        read -p \"Command failed to run. Press [Enter] to exit.\"\n"
         << "        exit 1\n"
         << "    fi\n"
         << "}\n"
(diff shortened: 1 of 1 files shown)


src/fabric/pyvenvmanager.cpp
```diff
@@ -297,7 +297,7 @@ auto createPythonVirtualEnv(const QString &venvName, const QString &requirements
         << "    if [ $? -ne 0 ]\n"
         << "    then\n"
         << "        echo \"\"\n"
-        << "        read -p \"Command failed to run. Press enter to exit.\"\n"
+        << "        read -p \"Command failed to run. Press Enter to exit.\"\n"
```
ximion:
```suggestion
        << "        read -p \"Command failed to run. Press [Enter] to exit.\"\n"
```

src/fabric/pyvenvmanager.cpp
```diff
@@ -310,7 +310,7 @@ auto createPythonVirtualEnv(const QString &venvName, const QString &requirements
         << "run_check pip install -r " << shellQuote(tmpRequirementsFname) << "\n"
 
         << "echo \"\"\n"
-        << "read -p \"Success! Press any key to exit.\""
+        << "read -p \"Success! Press Enter to exit.\""
```
ximion:
```suggestion
        << "read -p \"Success! Press [Enter] to exit.\""
```

ximion (commented):
Just a minor suggestion to make sure it's indicated that this is a key-command (putting it in quotemarks would also work, or we could do something really fancy and use colors - but square brackets are good enough, for now).

PR #184, opened 2026-07-17
VAAPI fixes

caenrigen:
I was playing around with the hardware encoding and decided to fix a few things:

- Surfaces return error code from sending frames to AV1 encoder (related issue I was debugging will follow in separate PR)
- Fixes the quality parameter not being passed into VAAPI, Codex suggested this fix that seems to actually cover all the codecs at once.

```
[av1_vaapi @ 0x7f5c67530400] No quality level set; using default (25).

[vp9_vaapi @ 0x7f5c5c4a0400] No quality level set; using default (100).

[h264_vaapi @ 0x7f5c67537800] No quality level set; using default (20).
```

---

Ironically the VAAPI "acceleration" seems to be slower than software encoding, expected? I am using an integrated `Arrow Lake-S [Intel Graphics]` (the only one compatible with VAAPI from the dropdown)

### DIFF
--- a/modules/videorecorder/videowriter.cpp
+++ b/modules/videorecorder/videowriter.cpp
@@ -797,16 +797,31 @@ void VideoWriter::initializeInternal()
 
     // set bitrate/crf
     d->cctx->bit_rate = 0;
-    av_dict_set_int(&codecopts, "crf", 0, 0);
-    if (d->codecProps.mode() == CodecProperties::ConstantQuality)
-        av_dict_set_int(&codecopts, "crf", d->codecProps.quality(), 0);
-    else if (d->codecProps.mode() == CodecProperties::ConstantBitrate)
-        d->cctx->bit_rate = d->codecProps.bitrateKbps() * 1000;
 
-    if (d->codecProps.useVaapi()) {
-        // some hardware-accelerated codecs use different options for some reason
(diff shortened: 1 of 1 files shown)


ximion:
> Ironically the VAAPI "acceleration" seems to be slower than software encoding, expected? I am using an integrated `Arrow Lake-S [Intel Graphics]` (the only one compatible with VAAPI from the dropdown)

That depends on your CPU - SVT-AV1 is quite fast if it has access to AVX512 instructions, and your CPU might be very capable while your GPU is on the low end, so just transferring the buffers around may slow things.
My RX 7900 XT beats the CPU with absolute ease though.

Have you tested that the `d->cctx->global_quality` change actually works? I remember this to be incredibly annoying to get right and work with hardware encoders, but that was also with FFmpeg 4.x, and it is very possible that newer versions handle this a lot better.

caenrigen:
> Have you tested that the d->cctx->global_quality change actually works?

It does prevent these av1_vaapi/vp9_vaapi/h264_vaapi warnings from being logged out, so the value is being picked up at least in some way. But I did not explore the effect of changing the quality and verifying the output in more detail.

btw, setting the constant quality to `0` will make the encoders use some internal default it seems, whose value I suppose won't be save in the module settings. Intended behaviour?

caenrigen:
Made some changes after some more ping pong with Codex

ximion:
The recent changes in master all include the changes made here already, so I am closing this. Thank you for the patches!

PR #185, opened 2026-07-17
videowriter: Fix AV1 VAAPI

caenrigen:
before this fix:

```
19:40:35.607890 E videorecorde-1:mod.videorecorder-0: Error raised by 'Video Recorder':
Unable to send frame 1 to encoder: Resource temporarily unavailable.
```

(Error code message already fixed via #184)

Codex says:

> this points to  an encoder-loop bug, Syntalos is not following FFmpeg’s send/receive backpressure contract:
> Syntalos calls avcodec_send_frame().
> If that returns EAGAIN, Syntalos currently treats it as fatal.
> Correct behavior is: call avcodec_receive_packet() in a loop until it returns EAGAIN, write any packets, then retry avcodec_send_frame() with the same frame.

I have no idea if Codex is right or not, but it indeed solved the issue, AV1 via VAAPI seems to work. I did not yet look at an actual recording contents to double check it looks good.

### DIFF
--- a/modules/videorecorder/videowriter.cpp
+++ b/modules/videorecorder/videowriter.cpp
@@ -443,7 +443,6 @@ public:
 
         hwDevCtx = nullptr;
         hwFrameCtx = nullptr;
-        hwFrame = nullptr;
 
         selectedEncoderName = QStringLiteral("No encoder selected yet");
     }
@@ -488,7 +487,6 @@ public:
 
     AVBufferRef *hwDevCtx;
     AVBufferRef *hwFrameCtx;
-    AVFrame *hwFrame;
 };
 #pragma GCC diagnostic pop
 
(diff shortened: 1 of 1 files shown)


ximion:
Please look at what is actually recorded! Hardware encoding works, but so far has only been properly tested with VP9 and HEVC, as AV1 encoding hardware was quite hard to get for a long time.

caenrigen:
> Please look at what is actually recorded!

Video looks fine 🙂 Visually seems the same as the av1 software encoded.

How to tweak the quality and how to compare it to software encoded I have no idea (if possible at all). Lmk if there some additional testing I could do to confirm it works as expected. In any case, that is a slightly different topic compared to what this PR attempts to fix (packets draining/encoder loop).

caenrigen:
@coderabbitai review

caenrigen:
> Please look at what is actually recorded! Hardware encoding works, but so far has only been properly tested with VP9 and HEVC, as AV1 encoding hardware was quite hard to get for a long time.

Actually after some more Codex ping pong, video inspection etc., it turns out there is another bug in there that (I think) was overriding frames that the encoder used as reference, leading to wrong output (though the file size is similar and the visual contents might be tricky to spot).

ximion:
I couldn't reproduce this on my machine, but I *could* find encoding issues when using H.264 specifically (not with any other codec) and your patch fixes that and looks overall correct too.

This should be safe, but please test it!

PR #186, opened 2026-07-17
videowriter: Prevent silently ignoring AV1 unused codec options

caenrigen:
lossless=1 has been silently ignored in AV1 and created the confusion I mentioned in https://github.com/syntalos/syntalos-web/pull/14

this patch does not fix that, but at least yields a warning about unused AV1 encoder options

for software encoding AV1 ignores `lossless=1` and for hardware encoding `crf=`:

```
22:38:55.517082 W videorecorde-1:mod.videorecorder-0: Video encoder ignored codec option(s): lossless=1
22:38:55.519881 W videorecorde-2:mod.videorecorder-1: Video encoder ignored codec option(s): crf=63
```

idk if lossless being ignored is due to the libraries on my machine or if it generally not available and should be simply disabled (or if it requires some other parameter e.g. `qp=0`)

### DIFF
--- a/modules/videorecorder/videowriter.cpp
+++ b/modules/videorecorder/videowriter.cpp
@@ -918,6 +918,21 @@ void VideoWriter::initializeInternal()
         throw std::runtime_error(
             std::format("Failed to open video encoder with the current parameters: {}", averrorToString(ret)));
     }
+    if (codecopts != nullptr) {
+        const AVDictionaryEntry *entry = nullptr;
+        std::string unusedCodecOpts;
+        while ((entry = av_dict_get(codecopts, "", entry, AV_DICT_IGNORE_SUFFIX)) != nullptr) {
+            if (!unusedCodecOpts.empty())
+                unusedCodecOpts += ", ";
+            unusedCodecOpts += entry->key;
+            unusedCodecOpts += "=";
+            unusedCodecOpts += entry->value;
+        }
+        if (!unusedCodecOpts.empty())
+            LOG_WARNING(d->log, "Video encoder ignored codec option(s): {}", unusedCodecOpts);
+
+        av_dict_free(&codecopts);
+    }
 
     // stream codec parameters must be set after opening the encoder
     avcodec_parameters_from_context(d->vstrm->codecpar, d->cctx);


ximion:
Looks like this has changed quite a bit from when I last touched it... SVT-AV1 does support lossless encoding, but at least on the CLI, the options are passed as their own special options block via `svtav1-params`: https://www.ffmpeg.org/ffmpeg-all.html#libsvtav1

We may need to move them there in the API calls as well, so options take effect. All tunables are documented at https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Docs/Parameters.md

We have had multiple breaking-changes releases since this was implemented as experimental feature, so, would make sense to tune it to current SVT-AV1 (which is still the best and fastest production AV1 encoder).

caenrigen:
Fun fact, I recorded some acquisitions in "lossless AV1" mode and have been in thinking that AV1 was magically losslessly compressing my videos 😂 while in reality it was just ignoring the flag and using `crf=0` 😂😂😂

impressive compression nonetheless I guess since the image quality seemed very good anyway

I need to track surgical instruments (potentially a needle too), so I will have to play around with compression quality options 😅

caenrigen:
> Looks like this has changed quite a bit from when I last touched it... SVT-AV1 does support lossless encoding, but at least on the CLI, the options are passed as their own special options block via `svtav1-params`: https://www.ffmpeg.org/ffmpeg-all.html#libsvtav1
> 
> We may need to move them there in the API calls as well, so options take effect. All tunables are documented at https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Docs/Parameters.md
> 
> We have had multiple breaking-changes releases since this was implemented as experimental feature, so, would make sense to tune it to current SVT-AV1 (which is still the best and fastest production AV1 encoder).

Glad to hear there has been some good progress!

Feel free to drop this PR if needed 🙂

ximion:
> impressive compression nonetheless I guess since the image quality seemed very good anyway

AV1 is truly incredible, if you have a beefy computer or dedicated hardware to perform the encoding, it is by far the best option for video encoding. On a slow system or older CPU (if you don't have AVX2, forget about it) it's unusable for encoding though.

caenrigen:
I indeed have a beefy computer at work. Also great for compiling iox2 and Syntalos pretty fast 😁

Btw, would be open to a PR that adds NVIDIA NVEC support? I have an RTX 5090 that can encode/decode AV1

Syntalos detects my RTX 5090 in the VAAPI list and then it fails (ofc as expected since it is not a VAAPI compatible device, potential improvement would be to check first for VAAPI compatibility)

ximion:
> Btw, would be open to a PR that adds NVIDIA NVEC support?

I wouldn't be too excited about it as it would add an extremely hard to test additional codepath that only works for one specific GPU vendor and that will essentially be abandoned in the code once committed. I was hoping that encoding support would eventually end up in Nvidia-VAAPI, but that's still WIP: https://github.com/elFarto/nvidia-vaapi-driver/pull/427

Nowadays, I think the thing that would make more sense to support for the future is Vulkan Video and Vulkan Compute Shaders for video encoding: https://developer.nvidia.com/vulkan/video/get-started / https://www.khronos.org/blog/video-encoding-and-decoding-with-vulkan-compute-shaders-in-ffmpeg

Those will also not introduce a dedicated codepath for one GPU vendor, and maybe can eventually (long-term) even replace VAAPI if the Vulkan API is flexible enough.

> Syntalos detects my RTX 5090 in the VAAPI list and then it fails

It is so annoying that Nvidia had to roll its own thing with VDPAU and then nvenc/nvdec. Granted, for the former they were very early, but it would have been nice if they would just have worked on a common API as well... Anyway, it looks like Nvidia-VAAPI and Vulkan Video might solve this finally.

caenrigen:
> I wouldn't be too excited about it as it would add an extremely hard to test additional codepath that only works for one specific GPU vendor and that will essentially be abandoned in the code once committed.

I definitely understand! This was more of a curiosity, at least so far. My beefy CPU at work seems to handle AV1 encoding great 😁

The actual reason I started "poking" into the hardware encoding/decoding is that I run into a bug in `edlio` that seems to not be able to load my AV1-software encoded videos (while ffmpeg itself has no issues with it). I will report this separately once I debug it on Monday.
 
> It is so annoying that Nvidia had to roll its own thing with VDPAU and then nvenc/nvdec. Granted, for the former they were very early, but it would have been nice if they would just have worked on a common API as well... Anyway, it looks like Nvidia-VAAPI and Vulkan Video might solve this finally.

At least NVIDIA finally has open source drivers 😂

ximion:
> The actual reason I started "poking" into the hardware encoding/decoding is that I run into a bug in `edlio` that seems to not be able to load my AV1-software encoded videos (while ffmpeg itself has no issues with it). I will report this separately once I debug it on Monday.

That sounds strange, maybe the OpenCV-Python module that you are using hasn't been compiled with the right codecs. This is a known bug with the conda version of it, and it has been incredibly annoying for a while...

> At least NVIDIA finally has open source drivers 😂

Well... Partially. It has open-source kernel drivers that will never be merged into the kernel (they are not up to the kernel standards and reworking them to fit would be massive work - but they are GPL-compatible, which solves a large chunk of legal and engineering issues, as more API is now available to them). But their userspace drivers are still fully closed-source, and there has been no indication that that would change.

Things are definitely better, but not on-par with AMD/Intel yet. But also, Nvidia is industry-leading, so the pressure for changes is very low, now that the biggest painpoint (closed kernel driver component) is somewhat solved-ish...

caenrigen:
> Well... Partially. It has open-source kernel drivers that will never be merged into the kernel (they are not up to the kernel standards and reworking them to fit would be massive work - but they are GPL-compatible, which solves a large chunk of legal and engineering issues, as more API is now available to them). But their userspace drivers are still fully closed-source, and there has been no indication that that would change.
> 
> Things are definitely better, but not on-par with AMD/Intel yet. But also, Nvidia is industry-leading, so the pressure for changes is very low, now that the biggest painpoint (closed kernel driver component) is somewhat solved-ish...

I was not aware of the details, learnt something today 🎩

---

Regarding this PR, are you considering merging it or rather working on some more proper fix?

ximion:
I need some time to sit down with this and read recent documentation to understand the changes made, then test the output. Encoder changes from my experience can have very strange sideeffects depending on the version used, the hardware present and the video input, so I want this to be sane.

And currently I just have to prioritize some (also open source) tasks that I am hired for first (because being at a conference and then on vacation as a consultant means zero income at all, while having more expenses and paying running expenses at their "as-if-you-earn-money" rate, and Germany is really expensive for that - so I need to get on top of pending contract tasks now first). Then I will get back to this :-) (shouldn't be too long, but may take at least another week).

caenrigen:
Hey Matthias! Thank you for letting me know. No worries, take care of yourself first, of course! 🙂 Life in Central Europe is not that cheap sadly 🥲

My encoder-related issues and PRs are not a priority at all, specially the hardware-based encoding 😉

ximion:
I implemented a bit of refactoring, can you test if this also works for you? Especially VA-API can be a bit weird depending on the hardware :-)

caenrigen:
Seems like you brewed some awesome refactoring, thank you very much for your efforts! 😄😄😄

---

NB Debian stable is on older AV1 so it emits a warning/error:

```
Svt[info]: SVT [version]:       SVT-AV1 Encoder Lib v2.3.0
Svt[info]: SVT [build]  :       GCC 14.2.0       64 bit
Svt[info]: -------------------------------------------
[libsvtav1 @ 0x7ff735700400] Error parsing option lossless: 1.
```

I suppose this is not intentional and should fail the run instead since the encoder cannot honor the `lossless` mode the GUI promised

PR #189, opened 2026-07-26
Camera arv fix buffer ownership

### LINKED ISSUE #188: camera-arv: Frame buffer ownership issues leads to corrupted recordings (e.g. consecutive duplicate frames)
It seems that `camera-arv`:
 
- Decodes into reusable internal storage.
- Obtains a shallow cv::Mat reference.
- Pushes it downstream without cloning the pixels.

Occasionally consecutive frames are corrupted, since the image in front of the camera does _not_ usually change dramatically from frame to frame it has not been trivial to spot!

Pretty nasty, specially for experiments that rely on pixel-perfect recordings 😅

The issues has been around since 78c5e6e9

PR with test and potential fix will follow up shortly.

I spotted the problem myself and Codex helped with the rest

caenrigen:
Fixes #188 

Adds a test that proves the duplicated-frames corruption and implements the necessary fix.

Codex implemented the test.

The test is perhaps too elaborate and consumes another ~10s or so, worth keeping? up to you 🙂

On a more serious topic, this might have affected people using camera-arv in the last 2 years 😅 maybe worth a loud warning in the README and/or next release notes


ximion:
Ooof, this will destroy high-speed data acquisition cases, especially since this already copies data so much for decoding. But we do only get a reference to a recycled memory region, so there is no way around making a deep copy.

I do wonder if this allocates every time, if so, we can forget about any performance of this codepath, because not only do we allocate in a hot loop, but also free the allocated region on a different thread (that's basically guaranteed).
Last time in benchmarked this was before the use of mimalloc though, which optimizes exactly this case though, so maybe mimalloc saves us. Also, OpenCV may be smart enough to allocate a bigger pool and recycle memory, but I'll have to read up on that, it's been a while since I looked into this...

I do think though the thing we should do is to change the decoders to return a full cv::Mat and not a reference. That way, we save an entire memcpy for every frame (and one allocation at the start, although that one is negligible). Ironically, this will make the single-threaded preview case a lot more inefficient.

I'll need some time to think about this a bit more :-)

caenrigen:
### TL;DR

I pushed a decoder "API" change to `decodeInto()` caller-provided buffers.

Fixes the bug. No `clone()`. Preview remains efficient by reusing its buffer. About 2x speed up over `clone()` (can push Codex' mini benchmark if relevant)

I know you have other commitments, so come back to this when you have time. I can use this draft for now to work around the bug.

### Codex' replies to some of your comments

> Several ownership/allocation claims need qualification:
> 
> - The old code did obtain a shallow cv::Mat alias. It was not a C++ reference, but it shared the decoder’s pixels without copying.
> - The published frame did not have exclusive ownership. For SwScaleDecoder, it did not own the pixels at all.
> - There is no OpenCV CPU cv::Mat pool recycling these frame allocations. Syntalos’ mimalloc allocator may recycle memory internally.
> - Changing the decoder to produce an independently owned cv::Mat directly is the correct next optimization—but merely changing the return type is insufficient because getCvImage() already returns by value.

> “It does not obtain a shallow cv::Mat reference”: Technically incorrect. OpenCV copying/assignment copies only the header and increments the refcount; it is an O(1) shallow copy. The installed OpenCV documentation says exactly that in /usr/include/opencv4/opencv2/core/mat.hpp:667.

> “It owns the memory region”: Only partially. Mono/Bayer frames share ownership with the decoder. They do not own an independent snapshot. SwScaleDecoder constructs a Mat around FFmpeg memory, so its Mat has no ownership/refcount at all: swscaledecoder.cpp (line 114).

> “A deep copy is unavoidable”: Correct under the current decoder contract. Not fundamentally unavoidable: direct decode into an independently owned output or a lifetime-managed pool removes the post-decode copy.

> “clone() allocates every frame”: Correct logically. OpenCV clone() creates an empty destination and copies into it. Syntalos’ allocator calls mi_malloc_aligned and constructs a new UMatData: syntalos/src/symemopt.cpp:83.

> “OpenCV may pool/recycle it”: Incorrect for CPU Mats. OpenCV’s standard allocator directly allocates/frees each Mat; Syntalos replaces that with an equivalent mimalloc-backed allocator. See OpenCV’s allocator implementation.

> “It is freed on another thread”: Usually true here, not guaranteed. With one downstream subscriber, the last reference normally dies on the consumer thread; with multiple subscribers, whichever releases last frees it.

> “mimalloc may save us”: It helps substantially with allocation and cross-thread deallocation. Mimalloc explicitly has separate concurrent free lists, making many remote frees a single CAS. It cannot remove the full-frame memcpy or its memory-bandwidth/cache cost. Official mimalloc design notes.

> “Make decoders return a full cv::Mat”: Directionally correct. But `virtual const cv::Mat getCvImage()` already returns by value. The API must instead produce a new owning allocation or write into a caller-owned destination. Removing const alone changes nothing.

> A further important detail: OpenCV’s Mat::create() reuses an existing destination whenever dimensions and type match—it does not check whether the pixels are shared. Therefore Bayer’s cvtColor(..., decoded, ...) continues overwriting queued frames even though downstream Mats increased the refcount.

### Codex' further optimization suggestions

#### Move decoding off the Aravis receiving thread

> Currently rawFrameCallback() invokes the complete decoder, transforms, clone and stream push synchronously: syntalos/modules/camera-arv/qarv/qarvcamera.cpp:433.
> Aravis explicitly warns that this callback runs on its receiving thread and that CPU-intensive processing should happen elsewhere. [Aravis callback documentation](https://aravisproject.github.io/aravis/aravis-stable/callback.StreamCallback.html).
> A higher-performance architecture would:
> Pop the completed ArvBuffer.
> Place it in an SPSC queue for the module’s dedicated worker.
> Return from the receive callback quickly.
> Decode on the worker.
> Requeue the Aravis buffer after decoding.
> arv_stream_push_buffer() and the pop operations are documented as thread-safe, so requeueing from the worker is supported. [Aravis thread-safety documentation](https://aravisproject.github.io/docs/aravis-0.8/thread-safety.html).
> This improves capture robustness independently of the ownership fix.

#### Add an adaptive decoded-frame pool only if profiling warrants it

> For no normal-path allocation:
> Prewarm several exact-size Mats.
> Decode directly into an acquired Mat.
> Return its storage to the pool only when the last downstream cv::Mat reference disappears.
> Grow on demand rather than blocking or dropping when the pool is exhausted.
> Use mimalloc as the fallback.
> A module-local cv::MatAllocator can return storage to a thread-safe free list from its deallocate() callback. This preserves normal Mat ownership semantics. A fixed round-robin ring is unsafe because consumer latency is unbounded.
> Do not inspect Mat::u->refcount to decide whether reuse is safe; OpenCV does not expose a supported thread-safe uniqueness API. 

#### Optional true zero-copy paths

> For native layouts such as Mono8, Mono16 and BGR8 with no layout-changing transform, Syntalos could publish a Mat backed directly by the Aravis buffer and requeue that buffer when the last consumer releases it.
> This is the absolute fastest path, but it couples the capture-buffer pool to downstream latency and complicates shutdown/backpressure. RGB8, Bayer and YUV still need conversion, so it would be a specialized optimization rather than the general fix.
> For maximum recording throughput, a separate raw-camera output consumed directly by the recorder—with decoded preview on a throttled branch—would go further still.

ximion:
Man Codex is relentless - it didn't quite get at what I meant though and misinterpreted...
I'll have a look at this again.

ximion:
A few nitpicks for the test specifically, I will fix them post-merge.

ximion:
I piled some efficiency changes on top, this should actually work very nicely now :-) - But as usual, please test it!

PR #192, opened 2026-07-28
fabric: Release GL context when closing GUIs

caenrigen:
Today I run into a strange bug (likely surfaced while accessing my workstation remotely via NoMachine).

```
13:11:04.192318 W syntalos:default: QOpenGLWidget: Failed to make context current
13:11:04.205164 W syntalos:default: QOpenGLWidget: Failed to make context current
13:11:04.222719 W syntalos:default: QOpenGLWidget: Failed to make context current
13:11:04.239083 W syntalos:default: QOpenGLWidget: Failed to make context current
13:11:04.426912 W syntalos:default: QOpenGLWidget: Failed to create wrapper texture
13:11:04.426919 W syntalos:default: QOpenGLWidget: No fbo, cannot render
13:11:06.282908 W syntalos:default: QOpenGLWidget: Failed to make context current
...
14:14:51.232113 W syntalos:default: Unrecognized OpenGL version
14:14:51.232116 W syntalos:default: Unrecognized OpenGL version
```

**Video: https://youtu.be/roaXShVx9kA**

Issue: A QOpenGLWidget can leave its context current after its top-level window is hidden. On GLX, restoring that context later may target the old native drawable and prevent another module window from creating its GL context.

Currently this affects camera-arv, canvas, plot-timeseries and latencycheck modules.
This fix safeguards future GUIs too.

NB Codex implemented.

The alternative would be to call the `doneCurrent()` on a per-module basis.
Codex also pointed out that in some cases you could replace `QOpenGLWidget` with simple `QWidget()` to remove the GLX/FBO/context machinery completely.

### DIFF
--- a/src/fabric/moduleapi.cpp
+++ b/src/fabric/moduleapi.cpp
@@ -22,7 +22,9 @@
 
 #include "config.h"
 #include <cmath>
+#include <QEvent>
 #include <QGuiApplication>
+#include <QOpenGLContext>
 #include <QScreen>
 #include <QDir>
 #include <QMainWindow>
@@ -39,6 +41,43 @@
 #include "utils/style.h"
 
 using namespace Syntalos;
+
+namespace
+{
+
+void releaseCurrentOpenGLContext()
+{
+    // A QOpenGLWidget can leave its context current after its top-level window
+    // is hidden. On GLX, restoring that context later may target the old native
+    // drawable and prevent another module window from creating its GL context.
+    // Releasing the binding does not destroy the context or any GL resources.
+    if (auto *context = QOpenGLContext::currentContext())
+        context->doneCurrent();
+}
+
+class ModuleWindowCloseEventFilter final : public QObject
+{
+public:
+    explicit ModuleWindowCloseEventFilter(QObject *parent)
+        : QObject(parent)
+    {
+    }
+
+protected:
+    bool eventFilter(QObject *, QEvent *event) override
+    {
+        if (event->type() == QEvent::Close)
+            releaseCurrentOpenGLContext();
+        return false;
+    }
+};
+
+void watchModuleWindowClose(QWidget *window)
+{
+    window->installEventFilter(new ModuleWindowCloseEventFilter(window));
+}
+
+} // namespace
 
 QString Syntalos::connectionHeatToHumanString(ConnectionHeatLevel heat)
 {
@@ -884,14 +923,18 @@ bool AbstractModule::isSettingsUiVisible()
 
 void AbstractModule::hideDisplayUi()
 {
-    for (auto const &wp : d->displayWindows)
+    for (auto const &wp : d->displayWindows) {
+        releaseCurrentOpenGLContext();
         wp.first->hide();
+    }
 }
 
 void AbstractModule::hideSettingsUi()
 {
-    for (auto const &wp : d->settingsWindows)
+    for (auto const &wp : d->settingsWindows) {
+        releaseCurrentOpenGLContext();
         wp.first->hide();
+    }
 }
 
 void AbstractModule::serializeSettings(const QString &, QVariantHash &, QByteArray &)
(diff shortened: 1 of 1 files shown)


ximion:
> Issue: A QOpenGLWidget can leave its context current after its top-level window is hidden. On GLX, restoring that context later may target the old native drawable and prevent another module window from creating its GL context.

This is not how QOpenGLWidget actually works...
And this patch is totally blind and messes with whatever context is current. Can you reproduce this issue somehow? I think this may be something else entirely...

caenrigen:
> > Issue: A QOpenGLWidget can leave its context current after its top-level window is hidden. On GLX, restoring that context later may target the old native drawable and prevent another module window from creating its GL context.
> 
> This is not how QOpenGLWidget actually works... And this patch is totally blind and messes with whatever context is current. 

Can very well be that Codex or my interpretation is flat wrong 🙈 

> Can you reproduce this issue somehow? I think this may be something else entirely...

Did you watch my video? See what happens in the Preview at 0:16

> **Video: https://youtu.be/roaXShVx9kA**

caenrigen:
Just checked and I am still able to reproduce it on df24bfad6a542535763f8eafc94a12bad4b2c76b

NB 0834733a4bc2e263b9bef462935fa99beda653c5 broke the build

caenrigen:
I went back to the Codex chat that debugged and implemented this, this is what it said:

>  ## Diagnosis
> 
> This is a Qt/NVIDIA GLX context-lifecycle failure, not a camera or image-decoding problem.
> 
> ## What happens
> 
> 1. Each camera preview is a `QOpenGLWidget` (`modules/camera-arv/glvideowidget.h:33`), including the idle SVG placeholder.
> 2. Closing the first camera window destroys its X11 drawable while its GL context is still current.
> 3. When Qt creates the next window, it temporarily switches GLX contexts and then attempts to restore the now-invalid old drawable. The `glXMakeCurrent()` call fails.
> 4. Qt does not check that failure before querying `GL_VERSION`; it receives an empty value and prints `Unrecognized OpenGL version` twice. The duplicate messages are two checks failing on the same empty string. Qt’s GLX source shows this sequence.
> 5. Sometimes Qt recovers when binding the real window, producing the harmless two-warning case. Sometimes the failure persists long enough that the preview FBO cannot be created, producing:
> 
>    - `Failed to make context current`
>    - `Failed to create wrapper texture`
>    - `No fbo, cannot render`
> 
> `QOpenGLWidget` renders through an off-screen FBO. Without it, the preview rectangle contains stale backing-store/compositor pixels, which explains the apparent image of the GUI itself. This is not a recursive camera image or a corrupted camera buffer. Qt documents the FBO behavior.
> 
> ## Evidence
> 
> - I reproduced the exact two warnings with a minimal camera-free Qt 6.8.2 program on this machine.
> - GLX tracing confirmed the failed restore and the subsequent failed dummy-context bind.
> - Calling `doneCurrent()` on the first preview before closing eliminated every failure.
> 
> The current environment also looks sane:
> 
> - X11 `DISPLAY=:0`
> - Direct rendering enabled
> - NVIDIA RTX 5090
> - NVIDIA driver `595.71.05`
> - OpenGL 4.6 through NVIDIA GLX
> - No suspicious Mesa/GL override variables
> 
> NoMachine may influence timing or expose a driver/Qt bug, but it has not obviously poisoned the GL environment in this shell.
> 
> ## Proposed fix
> 
> The minimal Syntalos-side fix is likely to call `doneCurrent()` before accepting the close event at the end of `ArvConfigWindow::closeEvent()` (`modules/camera-arv/configwindow.cpp:1109`):
> 
> ```cpp
> videoWidget->doneCurrent();
> event->accept();
> ```
> 
> I verified this exact widget-level approach in the reproducer.
> 
> A more architectural alternative is to make this preview a normal `QWidget`, since its `paintGL()` only uses `QPainter` and no direct GL calls.
> 
> `QT_XCB_GL_INTEGRATION=xcb_egl` also avoided the bug, but it creates an OpenGL ES 3.2 context here and could conflict with desktop-GL shaders elsewhere in Syntalos. Therefore, the `doneCurrent()` fix is preferable.

Issue #194, opened 2026-07-28
crash collector bugs


caenrigen:
I have syntalos installed from source and during the last crash #193 I was not able to collect the dump info etc. It said nothing found

Codex says:

> Why the crash report is empty
> A real Syntalos core exists for PID 233973. Its DrKonqi record is there ~/.cache/drkonqi/crashes/syntalos.5f3f0c7b8df047119d295ca527b9e8e8.233973.1785267285000000.ini. The collector simply cannot see it.
> The collector only searches the systemd journal at syntalos/tools/crashreport/debugcollect.cpp:106. On this machine:
> The user is not in `adm` or `systemd-journal`.
> journalctl reports insufficient permissions.
> Syntalos logs only through a console sink; during this run stdout/stderr pointed to /dev/pts/2, so no application log was persisted.
> sd_journal_open() succeeds with an empty accessible view, so the collector mistakes “no permission/no accessible records” for “no crashes” at syntalos/tools/crashreport/crashreportdialog.cpp:144.
> That produces the supplied empty report.
> 
> There is also a separate collector crash: after saving the report, it segfaults during shutdown because it installs the Syntalos Qt/Quill logging handler but never calls shutdownSyLogSystem(). Qt emits a thread-storage warning during process teardown, which reaches an already-destroyed Quill logger manager. Three collector cores show the identical stack. Compare crash reporter syntalos/tools/crashreport/main.cpp:47 with the proper shutdown in Syntalos syntalos/src/main.cpp:190.
> The collector also uses startsWith("syntalos"), so it can accidentally select a syntalos-crashreport core instead of the application core.

NB Codex often runs sandboxed which sometimes can make it make mistakes about permissions etc..

caenrigen:
Update on why I am not able to collect crash dumps via Syntalos' GUI

In a nutshell, permission issues: my current user cannot read the journal entries.

The long story, however, is more soul crashing 😫😫😫

My machine is part of a local network of linux lab/server machines under a common domain (i.e. same linux users can log in on multiple machines as the same linux user, etc.). I switched to the domain user account on my workstation not too long ago, that is why I only run into this problem now.

[...]

Issue #193, opened 2026-07-28
libsvtav1 (svt-av1 v2.3.0) sometimes crashes syntalos when invalid parameters are used (e.g. odd size 257x257 frames)


caenrigen:
I run into a crash (the one I mentioned today in #77 ) that stems from svt-av1 (v2.3.0 on Debian Stable, no newer Backports available).
~50% reproducible.

I can live with this I think, fixing the crop window prevents the av1 encoders teardown. No idea if this could happen also during normal shutdowns. 

## TL;DR

Codex says

> SVT-AV1 2.3.0 maintains a process-global CPU-topology pointer named lp_group. Initialization parses /proc/cpuinfo and writes through that pointer without synchronization. Every encoder teardown frees and nulls the same pointer, regardless of other encoder instances.

## Detailed

> The core shows three overlapping recorder states:
> right-zoom rejects 257×256 with EINVAL and begins cleanup.
> A valid 2048×2048 right recorder is still initializing SVT-AV1.
> Another valid 256×256 recorder is waiting inside avcodec_open2().
> SVT-AV1 2.3.0 maintains a process-global CPU-topology pointer named lp_group. Initialization parses /proc/cpuinfo and writes through that pointer without synchronization. Every encoder teardown frees and nulls the same pointer, regardless of other encoder instances. See the official [topology initialization](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/v2.3.0/Source/Lib/Globals/enc_handle.c#L190-235) and [handle initialization/teardown](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/v2.3.0/Source/Lib/Globals/enc_handle.c#L2784-2853).
> Syntalos calls avcodec_free_context() after the odd-sized encoder fails at syntalos/modules/videorecorder/videowriter.cpp:929. That teardown nulls lp_group while the valid 2048×2048 encoder is using it. At the fault:
> lp_group == nullptr
> num_groups == 1
> SVT-AV1 dereferences the null pointer and segfaults.
> The supplied log shows the same sequence at invalid_enc_params.log:11.
> This explains the intermittency: if all valid recorder initializations finish before the invalid recorder unwinds, Syntalos catches the encoder error and stops gracefully. With more concurrent recorders, the overlap window becomes much larger.
> Recommended fix direction:
> Reject dimensions incompatible with the selected pixel format before calling avcodec_open2().
[...]

PR #195, opened 2026-07-28
trivial: Install missing libraries to fix crash collector not opening

caenrigen:
libsyntalos-fabric.so was not being installed during `ninja install`. Opening the crash collector requires it.

I run into this while trying to debug #193, sadly after this fix it could not read the dumps either, see #194

### DIFF
--- a/tools/crashreport/meson.build
+++ b/tools/crashreport/meson.build
@@ -52,5 +52,6 @@ executable('syntalos-crashreport',
     include_directories: [root_include_dir,
                           include_directories('../../src')],
     install_dir: get_option('libexecdir'),
+    install_rpath: sy_libdir,
     install: true
 )


ximion:
How very strange, this worked before... Quite possibly linking against a public library pulled in the right rpath, otherwise I can't explain why this was not an issue earlier.

Thanks for the fix!

caenrigen:
I used to run syntalos all the time from the build dir (because the issues were happening too often and it was easier).
Now I install from source all the time because I need more users to be able to run it. Maybe I just never tried before 🤷🏼‍♂️

ximion:
No, I know of labs which have used it just fine in installed form, so that's not it. Most likely the path was pulled in by another library, masking this issue until now.

ximion:
Ah, dang, forgot to change the commit message from `trivial: Install missing libraries` to `trivial: Add missing rpath for private libraries` because the former commit message is very misleading.
Well, at least it won't be confusing when writing the release notes (anything prefixed with `trivial` doesn't show up in these)

caenrigen:
Commit amend followed by force push on Syntalos master not an option? 🤷🏼‍♂️

caenrigen:
I was indeed not sure what to put in the commit message. Next time I will ask Codex to help with phrasing the commit message 🙈 at least at that it should be good 😂

ximion:
No, force-pushing to master is a big no-go and the branch is protected for that reason. Contributors and anyone who has forked/cloned the project need some reliance that stuff isn't just yanked away from under them (doing this would make all copies of the branch immediately conflict with the one in `origin` which is a nightmare to sort out).

PR #202, opened 2026-08-14
fabric: Fix nanoseconds timestamp

caenrigen:


src/fabric/utils/tomlutils.cpp
```diff
@@ -32,7 +32,7 @@ toml::time qTimeToToml(const QTime &qtime)
     ttime.hour = qtime.hour();
     ttime.minute = qtime.minute();
     ttime.second = qtime.second();
-    ttime.nanosecond = qtime.msec() * 1000;
+    ttime.nanosecond = qtime.msec() * 1000000;
```
ximion:
```suggestion
    ttime.nanosecond = qtime.msec() * US_PER_MS * NS_PER_US;
```

src/fabric/utils/tomlutils.cpp
ximion, line 27: Add `#include "datactl/syclock.h"` before logging.

ximion (changes requested):
Weird mistake... But exactly why we have the conversion helpers, so we should use them :-)

src/fabric/utils/tomlutils.cpp
caenrigen: Compiler requires `Syntalos::US_PER_MS * Syntalos::NS_PER_US`, should I commit that?

ximion: Yes, we can change the namespacing later in a different commit maybe, but this is fine to have for now and the least disruptive change :-)

caenrigen:
The rabbit caught the inverse conversion bug. Should be good to merge now i think

PR #204, opened 2026-08-20
videorecorder: Preserver FFV1 16 bits pixel format in recordings

caenrigen:
@ximion was this a typo? or is this intentional? if intentional it is not really lossless anymore 😅 

I am not using this format so far, Codex spotted it

### DIFF
--- a/modules/videorecorder/videowriter.cpp
+++ b/modules/videorecorder/videowriter.cpp
@@ -893,7 +893,7 @@ void VideoWriter::initializeInternal()
         if (d->inputPixFormat == AV_PIX_FMT_GRAY8)
             d->encPixFormat = AV_PIX_FMT_GRAY8;
         if (d->inputPixFormat == AV_PIX_FMT_GRAY16LE)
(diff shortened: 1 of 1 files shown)


caenrigen:
I saw the `// MKV apparently doesn't handle 16-bit gray` comment for the `Raw` encoding.

Quick search says "This capability is available through specific profiles: Matroska v4 and Matroska v3".

But if that is wrong, Syntalos should let users know their frames' quality is being downgraded. Perhaps a logger warning for minimally intrusive change

ximion:
Syntalos didn't support any 16-bit recording at all until quite recently ;-) So it wasn't an issue before changes were made that now make it possible for the pipeline to process 16-bit (arguably a much needed change).

Have you tested this change, or does it create corrupted MKV changes? I do recall that being the case, but that comment is *ancient* and we are on much newer versions of all libraries now, so this maybe can be dropped.

caenrigen:
I tested playing a recording both in the default debian player and VLC, it played good in both of them.

NB the Raw format code path needs fixing too besides what is in this PR

PR #207, opened 2026-08-28
Videotransform even-sized fixes

caenrigen:
Codex reported issues and fixes:

> **High — even-size crop correction can create an out-of-bounds ROI.**
> After constraining the ROI to the remaining image area, odd sizes are rounded down and values below two are forced back to two syntalos/modules/videotransform/videotransform.cpp:398, 
> syntalos/modules/videotransform/videotransform.cpp:405. The origin is not moved back inside the image.
> For a 640-wide image, x=639,width=1 becomes x=639,width=2. Processing it at syntalos/modules/videotransform/videotransform.cpp:319 throws an OpenCV ROI assertion. This is reachable through saved settings/headless operation and potentially the region selector.

> **Medium — odd-dimension handling is simultaneously too broad and incomplete.**
> The built-in Scale transform still rounds to arbitrary dimensions modules/videotransform/videotransform.cpp:484. Scaling 640×480 by 1.01 produces 646×485 and is rejected by the recorder. The commit’s “stop producing them” intention therefore only covers Crop.

it also says:

> modules/videorecorder/videowriter.cpp:1342 rejects every odd frame before considering the selected codec, pixel format, or whether the video is grayscale. FFV1 successfully encoded both grayscale and exact-color BGR0 at 641×479 in a direct FFmpeg check, so previously valid recordings now fail.

but I never really needed odd sizes, nor I see myself needing them anytime soon.

### DIFF
--- a/modules/videotransform/videotransform.cpp
+++ b/modules/videotransform/videotransform.cpp
@@ -487,19 +487,19 @@ void ScaleTransform::createSettingsUi(QWidget *parent)
 
 MetaSize ScaleTransform::resultSize()
 {
-    return {
-        static_cast<int>(std::round(m_originalSize.width * m_scaleFactor)),
-        static_cast<int>(std::round(m_originalSize.height * m_scaleFactor))};
+    const int width = std::max(2, static_cast<int>(std::round(m_originalSize.width * m_scaleFactor)));
+    const int height = std::max(2, static_cast<int>(std::round(m_originalSize.height * m_scaleFactor)));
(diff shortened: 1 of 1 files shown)


ximion:
> but I never really needed odd sizes, nor I see myself needing them anytime soon.

I genuinely think those are an antifeature, so many codecs either require them or work better with them that we should just make people do the right thing here and use even numbers. Adding/removing one line should not be a limiting factor for anything.

Thanks for the patch!

PR #208, opened 2026-08-28
Videorecorder enhancements and fixes: exact-color and gray16

caenrigen:
Not sure if you already had in mind to enhance/fix some of these things 🙈 

NB Mostly Codex-authored.

See commit messages for an overview (or wait for the rabbit hehe)

Let me know if you prefer me to drop any commits, etc..

NB commit messages are not 80-char word-wrapped.

### DIFF
--- a/modules/videorecorder/tests/test-videorecorder.cpp
+++ b/modules/videorecorder/tests/test-videorecorder.cpp
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2026 Matthias Klumpp <matthias@tenstral.net>
+ *
+ * Licensed under the GNU Lesser General Public License Version 3
+ */
+
+#include <QTemporaryDir>
+#include <QtTest>
+#include <cstring>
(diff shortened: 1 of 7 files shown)


ximion:
> See commit messages for an overview (or wait for the rabbit hehe)

I don't particularly trust the rabbit, although it has so far done a good job ;-)
I always look at the commit messages directly, reviewing patch-by-patch is also a lot easier than reviewing a big blob.

> NB commit messages are not 80-char word-wrapped.

Yeah, that's rather ugly - not a blocker, but please wrap them next time :-)

modules/videorecorder/tests/test-videorecorder.cpp
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (C) 2026 Matthias Klumpp <matthias@tenstral.net>
```
ximion, line 2: FWIW, there should be your name here, sometimes in addition to mine if you make substantial changes ;-)

PR #209, opened 2026-08-28
Read frame refcount atomically

caenrigen:
Codex seems to have identified a potential race condition in the recent datactl change.

See the test commit and its message with how it proved it. the other commit implements a solution but might be fragile and dependent on OpenCV internals.

Codex suggests as a long-term solution an explicit Syntalos buffer lease/pool whose slots are returned when the final Frame lease is released.

I understand too little of this, but since it seems to have a test that proves was worth reporting (i hope) 🙂

### DIFF
--- a/src/datactl/frametype.h
+++ b/src/datactl/frametype.h
@@ -47,8 +47,12 @@ namespace Syntalos
  */
 inline void matEnsureExclusive(cv::Mat &mat, int rows, int cols, int type)
 {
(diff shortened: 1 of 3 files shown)


ximion:
Codex is half-right. This is not a latent correctness bug, but we *should* use an atomic operation instead of reading the refcount with a plain load (not sure why I did that, probably it was test code that I didn't update).

The test is bogus, as far as I can tell, it's a test that can never fail. I implemented basically this solution, but with a tiny test that just checks that the current behavior is sane. Even before though, this would have never led to corruptions or issues, only to noise when running with sanitizers.

caenrigen:
Thank you, sir! 🎩

PR #210, opened 2026-09-01
datactl: Keep synchronized timestamps strictly increasing

caenrigen:
This is for the very unlikely edge case that the synchronizer would timestamp two consecutive frames with the same timestamp.

### DIFF
--- a/src/datactl/timesync.cpp
+++ b/src/datactl/timesync.cpp
@@ -709,8 +709,9 @@ void SecondaryClockSynchronizer::processTimestamp(
             // correct fluke unconditionally
             masterTimestamp = masterTimestampFAdj;
 
-            // prevent any time-travel into the past (this should be impossible, but better be safe)
-            if (masterTimestamp < m_lastMasterTS)
+            // Prevent any time-travel into the past (this should be impossible, but better be safe),
+            // and advance exact duplicates so adjusted timestamps remain strictly increasing.
+            if (masterTimestamp <= m_lastMasterTS)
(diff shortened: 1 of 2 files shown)


ximion:
Did you actually hit that case somehow and test this? The current behavior of not decreasing timestamps is very much intentional, as well as keeping the same timestamps in case we couldn't determine a correct one. Because if we were to just increase the timestamp, we would *also* be wrong, but in addition to being wrong, we would also make the issue completely undetectable.

So, increasing the timestamp would be "wrong + 1", so, arguably, be a little bit more wrong ;-) (but more importantly, nobody could notice these issues in data).
Ideally, this case is impossible to reach though anyway (which is why I wonder if you hit it - even with ultra high frequency data sources, this never happened during testing).

caenrigen:
I did not run into this. But I did not attempt hard either, nor I inpescted this in my few past dummy datasets.

One of my cameras might be able to do 500+ fps I think, can give it a try this week.

But indeed very unlikely. Feel free to close this.

caenrigen:
I did a few recordings with 750 FPS, no timestamp duplicates.

I also checked all the .tsync files for duplicates. Found one, old dummy/testing dataset, The timestamps were generated by my own Python module (the human eyes tracking glasses). It has been a while, might be due to older bugs.

Feel free to close this PR

ximion:
I would close this for now, as the current behavior has actually been really helpful to spot bugs in the past during analysis, and I wouldn't want to lose it :sweat_smile: 

This also is an edge case scenario that should never happen, but *if* it were to happen, we should be rather noisy about it and not hide it.

PR #211, opened 2026-09-01
fabric: Preserve subscription suspension on reset

### LINKED ISSUE #206: Memory explodes when streaming data from a Python Module into a disabled Module (again)
As [requested](https://github.com/syntalos/syntalos/issues/55#issuecomment-5450129524) new issue to track the problem.

This board explodes memory:

Image

Memory usage:

Image

It does reset when I stop the run so it is probably not a leak.

NB This was happening also before your recent refactor (I run into it before rebasing and then confirmed it happened after the rebase too)

caenrigen:
Fixes #206

NB Codex implemented test

### DIFF
--- a/src/fabric/streams/stream.h
+++ b/src/fabric/streams/stream.h
@@ -188,6 +188,7 @@ public:
           m_notifyPending(false),
           m_active(true),
           m_suspended(false),
+          m_dormant(false),
           m_throttle(0),
           m_skippedElements(0),
           m_log(getLogger("subscription"))
@@ -316,7 +317,7 @@ public:
 
     bool isSuspended() const
     {
-        return m_suspended;
+        return m_suspended || m_dormant;
     }
 
     /**
@@ -469,6 +470,7 @@ private:
     std::atomic_bool m_notifyPending;
     std::atomic_bool m_active;
     std::atomic_bool m_suspended;
+    std::atomic_bool m_dormant;
     std::atomic_uint m_throttle;
     std::atomic_uint m_skippedElements;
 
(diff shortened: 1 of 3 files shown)


caenrigen:
There seemed to be some more similar issues related to dormant/suspended logic, see last 4 commits (Codex implemented).

Maybe you prefer to solve things in some other ways etc.. force push changes as you see fit 🙂

ximion:
The patch doesn't look crazy, but this will take me a few days to think through (I need some time for that). I don't want to add even more state to streams, for example, "dormant" is a module property, while streams can either be active or suspended. So, a module being dormant might suspend the streams, but the streams don't need to know why they are suspended.

So, this patch feels like a layer violation, but depending on the context, maybe the streams do not have enough information to behave correctly and I do need to give them more context (if I do, I would do it as an enum though, instead of many booleans, because many of these states are mutually exclusive).
I have to think through this stuff though, the stream and module state machine became one of the most complex parts of the codebase over time (once disabling modules and OOP modules were added).

caenrigen:
I understand. Codex suggested doing some conceptual refactoring to disentangle some things. too many booleans becomes a code smell.

I will use codex to suggest something for inspiration maybe

ximion:
Codex was way off this time! The real issue and real solution was orders of magnitude simpler, without any changes to the state machine or layer violations. Just a plain old bug that was accidentally reintroduced. I fixed it for good now, and added a small sentinel that should catch a regression again in case it ever happens.

Please test this and see if the fix works for you!

caenrigen:
> Codex was way off this time! The real issue and real solution was orders of magnitude simpler, without any changes to the state machine or layer violations. Just a plain old bug that was accidentally reintroduced.

Upssss 🙈 Codex has a tendency for over engineering 😅 I saw some agent skills on GitHub aiming to reduce it.

> I fixed it for good now, and added a small sentinel that should catch a regression again in case it ever happens.

Awesome!

> Please test this and see if the fix works for you!

Solved it for me 😄

